PR #5, opened 2017-03-17
mongo input and refactor integration

pferrel:
changing the template interface to remove knowledge of event classes from the server, have Mongo input persistence working, simplified the Dataset interface and now endpoints for input and query both call Engine methods. The Engine is acting like the controller in the MVC sense.

The CBCmdLineDriver works but the ExampleEventClient or JavaSDK does not. The Java SDK has some needed changes.

@laser13 can you take a look? Look for `// Todo: Semen` comments and try running the `send_events.sh` 

When this is integrated we should push to develop, right? Feel free to leave comments on this PR I am not blocked by your changes. We need to resolve some of the Java SDK issues through because I need to document it for clients.

Keep in mind that we have bad events. They are legal events, just wrong for this template. This should cause a non-fatal error in the template validation code. Run CBCmdLineDriver to see what should happen. Here is my IntelliJ config:

image


pferrel:
@laser13 please review.

laser13 (approved):

java-sdk/pom.xml
```diff
@@ -29,20 +80,20 @@
     <dependencies>
         <dependency>
             <groupId>com.typesafe.akka</groupId>
-            <artifactId>akka-http_2.11</artifactId>
-            <version>10.0.3</version>
+            <artifactId>akka-http_2.12</artifactId>
+            <version>10.0.5</version>
```
pferrel, line 84: I thought we agreed to use Scala 2.11? There are several libs I need to use that aren't in 2.12.

PR #6, opened 2017-03-24
Training using Actors

pferrel:


laser13 (approved):

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -0,0 +1,184 @@
...
+    maxIter: Int = 100, // the rest of these are VW params
+    regParam: Double = 0.0,
+    stepSize: Double = 0.1,
+    bitPrecision: Int = 24,
+    modelName: String = "model.vw",
+    namespace: String = "n",
+    maxClasses: Int = 3)
+  extends AlgorithmParams
+
+class TrainerManager extends ActorWithLogging{
```
pferrel, line 80: @laser13 I was planning to have a collection of Actors for `SingleGroupTrainer`s that are managed by CBAlgorithm. 

Is a separate `TrainerManager` necessary? This seems to add a lot of complexity.

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -0,0 +1,184 @@
...
+      val exists = trainers.keys.toList
+      val diff = list.filter(!exists.contains(_))
+
+      log.info("Exists trainers: {}", exists)
+      log.info("New trainers: {}", list)
+      log.info("Diff trainers: {}", diff)
+
+      diff.foreach { case (trainer, collection) ⇒
+        val actor = context watch context.actorOf(SingleGroupTrainer.props(collection), trainer)
+        trainers += trainer → actor
```
pferrel, line 96: a Mutable Hashmap? Why not Immutable? There will never be a large number of `SingleGroupTrainers` per model.

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  def await: Receive = {
+    case Train ⇒
+      log.info(s"${name} Receive 'Run', run job")
+      log.info(s"${name} Switch context to work")
+      context become work
+      startWork() recover {
+        case ex: Throwable ⇒ log.error(ex, "ERROR!")
+      } map { _ ⇒
+        log.info(s"${name} Work complete!")
```
pferrel, line 142: Please no ⇒, can we stick with =>  ? I use too many different documentation tools that all have different ways to create "⇒" with some it is very hard.

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -0,0 +1,184 @@
...
+        context become await
+      }
+
+  }
+
+  def work: Receive = {
+    case Train ⇒
+      log.info(s"${name} Received 'Run', not react, because already started job")
+  }
+
```
pferrel, line 153: is this required? It does nothing, right?

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -0,0 +1,184 @@
...
+  override def preStart(): Unit = {
+    super.preStart()
+    log.info(s"${Console.GREEN}Start actor ${name}${Console.RESET}")
+  }
+
+  override def postStop(): Unit = {
+    super.postStop()
+    log.info(s"${Console.RED}Stop actor ${name}${Console.RESET}")
+  }
+
```
pferrel, line 183: Are `preStart` and `postStop` required? I assume they are here as an example of how to hook start and stop?

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -0,0 +1,184 @@
...
+      log.info(s"${name} Received 'Run', not react, because already started job")
+  }
+
+  private def startWork(): Future[Unit] = {
+    log.info(s"${name} Start work")
+    Future {
+      Thread.sleep(15000)
+    }
+  }
+
```
pferrel, line 160: Does this really have to return a Future? Since it is an actor I assumed that the work done as a result of any message could take any length of time without blocking any code execution? Can't this method just start training and return Unit when done? 

The `SingleGroupTrainer` never needs to return values, when it is done that DB will be modified and since there is only ever one per `Dataset` & group, this is sufficient.

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBDataset.scala
```diff
@@ -48,7 +52,8 @@ import com.mongodb.casbah.commons.conversions.scala._
   */
 class CBDataset(resourceId: String) extends Dataset[CBEvent](resourceId) {
 
-  lazy val store = new Mongo(m = "localhost", p = 27017, n = resourceId)
+  private lazy val config = ConfigFactory.load()
+  lazy val store = new Mongo(m = config.getString("mongo.host"), p = config.getInt("mongo.port"), n = resourceId)
   // Todo: need to get port from CLI or config
```
pferrel, line 57: Where are these being defined: "mongo.host", "mongo.port". Can we make them the default constructor params for `Mongo`?

pferrel: It looks like these are in a resource/application.conf. Not familiar with the way this works but assuming the values will come from environment variables but default to `localhost` `27017`?

These values needs to be user accessible and documented something like `pio-env.sh` so if I'm correct that application.conf look in the ENV for the Mongo settings, we should have a `aml.sh` that is sourced before launching the server, right?

pferrel (commented):
@laser13 I'm planning to work with this code but have several comments and questions, most regarding what is required vs example in the code.

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
laser13: No, it's just for the convenience of logging.

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBDataset.scala
laser13: Yes, MONGO_HOST & MONGO_PORT, these are the environment variables. By default localhost 27017

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
laser13: It is possible and without `Future`

PR #8, opened 2017-03-25
Python sdk

laser13:


pferrel:
@laser13 The REST side of commands and the Python CLI is one thing and the Python SDK is another. The Python SDK should have the same functionality as the Java SDK and is used as a client to the pio-kappa REST-server. 

At some point we may need to refactor the python into 2 separate things.

This way we have both a Java and Python SDK for client applications and a Python CLI, which has much code in common with the Python SDK, if not using it directly.

I don't have much time on my battery so will look through this more carefully before Monday.

Thanks, this looks like very good work.

rest-server/core/src/main/scala/com/actionml/core/validate/ParseError.scala
```diff
@@ -9,3 +9,4 @@ sealed trait ValidateError
 final case class ParseError(message: String) extends ValidateError
 final case class MissingParams(message: String) extends ValidateError
 final case class WrongParams(message: String) extends ValidateError
+final case class NotImplemented(message: String) extends ValidateError
```
pferrel: I assume we can use these in validating events and queries.

rest-server/src/main/scala/com/actionml/router/http/routes/BaseRouter.scala
```diff
@@ -42,12 +44,14 @@ abstract class BaseRouter(implicit inj: Injector) extends AkkaInjectable with Ci
 
   def completeByValidated(
     ifDefinedStatus: StatusCode
-  )(ifDefinedResource: Future[Validated[ValidateError, Json]]): Route =
+  )(ifDefinedResource: Future[Response]): Route =
     onSuccess(ifDefinedResource) {
```
pferrel, line 47: are all responses async Futures? Shouldn't some be forced to respond in real time? The event input and query should respond immediately or you should get a connection timeout, right?

Most commands are synchronous too.

Not sure if this is the right place to make this comment, but its true anyway :-)

rest-server/src/main/scala/com/actionml/router/http/routes/BaseRouter.scala
```diff
@@ -42,12 +44,14 @@ abstract class BaseRouter(implicit inj: Injector) extends AkkaInjectable with Ci
...
       case Valid(json) => complete(ifDefinedStatus, json)
       case Invalid(error: ParseError) => complete(StatusCodes.BadRequest, error.message)
       case Invalid(error: MissingParams) => complete(StatusCodes.BadRequest, error.message)
       case Invalid(error: WrongParams) => complete(StatusCodes.BadRequest, error.message)
+      case Invalid(error: NotImplemented) ⇒ complete(StatusCodes.NotImplemented, error.message)
+      case _ ⇒ complete(StatusCodes.NotFound)
```
pferrel: Good: `NotImplemented`

python-sdk/actionml/__init__.py
```diff
@@ -0,0 +1,634 @@
...
+        """
+        Asynchronously get an event from PIO Kappa Server.
+        :param event_id: event id returned by the EventServer when creating the event.
+        :returns: AsyncRequest object.
+        """
+        request = AsyncRequest("GET", self._add_segment(event_id))
+        request.set_response_handler(self._ok_response_handler)
+        self._connection.make_request(request)
+        return request
+
```
pferrel: This will not be implemented in v0.1.0 of pio-kappa

Many of these will not be implemented (though they will eventually) Stubbing them as not implemented if OK for now.

Only the CLI mentioned in the admin docs CLI should be implemented fully.

python-sdk/actionml/__init__.py
```diff
@@ -0,0 +1,634 @@
...
+    """
+        request = AsyncRequest("DELETE", self._add_segment(event_id))
+        request.set_response_handler(self._delete_response_handler)
+        self._connection.make_request(request)
+        return request
+
+    def delete(self, event_id):
+        """Synchronously delete an event from PIO Kappa Server."""
+        return self.async_delete(event_id).get_response()
+
```
pferrel, line 257: This will not be implemented in v0.1.0 of pio-kappa. No event manipulation except `input` is implemented for pipo-kappa-v0.1.0 the rest should report "not implemented"

Many of these will not be implemented (though they will eventually) Stubbing them as not implemented if OK for now.

Only the CLI mentioned in the admin docs CLI should be implemented fully.

python-sdk/actionml/__init__.py
```diff
@@ -0,0 +1,634 @@
...
+        Delete a user.
+        Wrapper of async_create function, setting event to "$delete" and entity_type
+        to "user".
+        """
+        return self.async_create(
+            event="$delete",
+            entity_type="user",
+            entity_id=uid,
+            event_time=event_time)
+
```
pferrel: Stick to REST defined API, we will never have a "user delete" because notion of a user is Template specific. Only resources defined in REST can be manipulated and the admin CLI only accesses these, dataset, engine, command.

These are the same rules as the Java SDK with the addition of the admin CLI + admin REST API. For instance queries are json strings, the structure is Template specific and not defined in the Python SDK or Python CLI.

pferrel: Another way to say this is that nothing outside the Template should define `async_delete_user` or any other Template specific object.

python-sdk/actionml/__init__.py
```diff
@@ -0,0 +1,634 @@
...
+    def delete_user(self, uid, event_time=None):
+        """Delete a user."""
+        return self.async_delete_user(uid, event_time).get_response()
+
```
pferrel: same as above

python-sdk/actionml/__init__.py
```diff
@@ -0,0 +1,634 @@
...
+    def async_set_item(self, iid, properties=None, event_time=None):
+        """
```
pferrel: same as above

and so on...

python-sdk/actionml/__init__.py
```diff
@@ -0,0 +1,634 @@
...
+            properties=properties,
+            event_time=event_time)
+
+    def record_user_action_on_item(self, action, uid, iid, properties=None, event_time=None):
+        """Create a user-to-item action."""
+        if properties is None:
+            properties = {}
+        return self.async_record_user_action_on_item(
+            action, uid, iid, properties, event_time).get_response()
+
```
pferrel: Most of these should be replaced with event construction like the Java SDK to keep this code decoupled from specific Templates

python-sdk/actionml/__init__.py
```diff
@@ -0,0 +1,634 @@
...
+    def async_delete(self, dataset_id: str):
+        request = AsyncRequest("DELETE", self._add_segment(dataset_id))
+        request.set_response_handler(self._delete_response_handler)
+        self._connection.make_request(request)
+        return request
+
+    def delete(self, dataset_id: str):
+        return self.async_delete(dataset_id).get_response()
+
```
pferrel, line 304: Datasets are defined in the REST API and in the admin CLI so these are good. 

We may not implement all, like GET /dataset/dataset-id for instance, and can report "not implemented" But they should be stubbed so we can implement in the future.

python-sdk/actionml/__init__.py
```diff
@@ -0,0 +1,634 @@
...
+    def async_delete(self, engine_id: str):
+        request = AsyncRequest("DELETE", self._add_segment(engine_id))
+        request.set_response_handler(self._delete_response_handler)
+        self._connection.make_request(request)
+        return request
+
+    def delete(self, engine_id: str):
+        return self.async_delete(engine_id).get_response()
+
```
pferrel, line 351: Similar comments to Dataset. Correct to have these, but only the methods required by the REST API or admin CLI may be implemented, the rest at "not implemented"

python-sdk/deploy.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+python3 setup.py register -r pypitest
+python3 setup.py sdist upload -r pypitest
```
pferrel, line 4: Python3 like Scala 12 may be problematic.

On Ubuntu you call python3 "python3" on other 'nix (like Mac BSD Unix) it replaces the default "python". How do we handle this so that it will run on any machine? Perhaps some kind of aliasing setup in the ENV before any of these are called? Can you propose something that will work on all important machines, Debian derived, Centos derived, Mac?

rest-server/src/main/scala/com/actionml/router/service/EngineService.scala
```diff
@@ -12,6 +15,30 @@ import scaldi.Injector
 
 trait EngineService extends ActorInjectable
 
+class CBEngineService(implicit inj: Injector) extends EngineService{
+
```
pferrel: The next version of pio-kappa should allow runtime addition of Templates so these Services will be configured by some parameters. Do nothing about this now, just FYI.

Some Datasets will invoke code from one Template and others code from another Template. These will always communicate with the "controller" which extends "Engine" but the instance may be different depending on the Template and the REST will be the same, though events will mean different things to different Templates and the same for queries. The REST will remain the same, what REST invokes will be different per Template.

Again, FYI only. Do nothing about this now.

pferrel (commented):
Good, let's not integrate with "integration" until after this week, because I want to deploy pio-kappa for a client and we will need to use this to work out some specifics. If you make changes, do so in the "python-sdk" branch

Thanks

rest-server/src/main/scala/com/actionml/router/http/routes/BaseRouter.scala
laser13: Here everything is correct
The server must be asynchronous

python-sdk/actionml/__init__.py
laser13: ok

python-sdk/actionml/__init__.py
laser13: Ok,  I'll delete it.

python-sdk/actionml/__init__.py
laser13: ok

pferrel: @laser13 I was wrong above, all `/dataset` endpoints are removed. we moved `/datasets/resource-id/events` to `/engines/resource-id/events` creating and deleting datasets is not supported in any way. We do it through creating and removing engines. 

We need some way to mirror the input json into HDFS files for later playback because once it is parsed and validated it is no longer an immutable event stream. This is best done in the REST-server by saving all raw json to some HDFS location then also sending it to the `engine.input` also. We should turn this on and off via `engine.json` params see: https://github.com/actionml/pio-kappa/blob/admin/pio-kappa-refactoring.md

python-sdk/actionml/__init__.py
pferrel: I was wrong above, all `/datasets` endpoints should be removed

python-sdk/actionml/__init__.py
laser13: @pferrel ОК, we can use https://flume.apache.org/FlumeUserGuide.html

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -32,60 +39,87 @@ import scala.concurrent.Future
...
-    diff.foreach { case (trainer, collection) ⇒
-      val actor = system.actorOf(SingleGroupTrainer.props(collection), trainer)
-      trainers += trainer → actor
-    }
-    this
+  def init(json: String): Validated[ValidateError, Boolean] = {
+    val response = parseAndValidate[CBAlgoParams](json)
+
+    if (response.isValid) {
+      params = response.getOrElse(CBAlgoParams())
```
pferrel, line 55: should use "andThen" here

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -32,60 +39,87 @@ import scala.concurrent.Future
...
-    trainers(groupName) ! SingleGroupTrainer.Train
+    try {
+      logger.trace("Train trainer {}", groupName)
+      trainers(groupName) ! SingleGroupTrainer.Train
+    } catch{
+      case e: NoSuchElementException =>
+        logger.error(s"Training triggered on non-existent group: $groupName The group must be initialized first. " +
+          s"All events for this group will be ignored. ")
+    }
   }
```
pferrel, line 81: using exceptions for existence check is bad form IMO, check the keyset of the Map instead.

pferrel: I'll switch to keyset check

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala
```diff
@@ -32,60 +39,87 @@ import scala.concurrent.Future
...
+      val actor = actors.actorOf(SingleGroupTrainer.props(collection), groupName)
       trainers += groupName → actor
     }
   }
 
-  def stop(): Future[Terminated] = {
-    system.terminate()
+  override def stop(): Unit = { // Todo: Semen, do we have to return Future[Terminated]? What is the benefit?
+    actors.terminate().wait() // Semen: I added the wait, not sure why we were returning a terminated Future
   }
```
pferrel, line 105: @laser13, is there are reason not to wait here?

pferrel (approved):
A few minor changes but I'll merge first

PR #10, opened 2017-04-29
Python cli to develop

pferrel:
@laser13 is this ready to review?


laser13:
@pferrel Now ready

https://github.com/actionml/pio-kappa/blob/python-cli/rest-server/install.md

pferrel:
@laser13 travis failure, may be building the maven artifact.

BTW We need an artifact for the Java SDK that contains all transitive dependencies. This is to make it easier to use Ant for building the SDK into a client.

PR #12, opened 2017-05-25
json mirroring

snovgorodtsev-zz:


pferrel:
can you write a paragraph describing the layout of the mirrored files with directory and file names?

pferrel:
@laser13 can you take a quick look at this PR?

pferrel:
To finish this we'll add a new parameter to the engine json params. Part of the params are read by Harness, part by the specific engine. We'll use the Harness part to specify the type and root location of mirroring. The config will be an `Option` so any part of it can be omitted and we'll use the default values you have already used.

{
    "engineId": "some-id",
    "engineFactory": "some-factory-classname",
    "mirroringType": "hdfs" | "localfs",
    "mirrorLocation": "path-or-descriptor-or-root-location"
    "algorithm": {}
}


@snovgorodtsev I'll create this change and some examples of how to read the location, you can take it from there and get it working. To do this we'll need to share a feature branch in git so I'll create one and share it with you, it will be a new PR to `develop`

Let me know when you have the Harness running and have questions.

laser13 (approved):

pferrel:
This work should be moved to https://github.com/actionml/harness/pull/16

PR #15, opened 2017-06-16
first changes to do mirroring

pferrel:
@snovgorodtsev pull from my personal branch and we can merge your changes with mine so the merge will be testable by both of us in the mirror branch


pferrel:
@snovgorodtsev look for comments with "Slava" these point out where you can get information about setting up mirroring

Like this one: https://github.com/pferrel/harness/blob/3cea414d39e566bbbf184beaacfe37274d6047c7/rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala#L35

and this one: https://github.com/pferrel/harness/blob/3cea414d39e566bbbf184beaacfe37274d6047c7/rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala#L43

PR #16, opened 2017-06-16
Mirror implementation

pferrel:
Slava, this does not compile yet. I tried to merge but there are some fundamental changes to the core API that I don't think we can accept for reasons we'll discuss in review


rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -3,7 +3,7 @@ package com.actionml.admin
 import cats.data.Validated
 import cats.data.Validated.{Invalid, Valid}
 import com.actionml.core.storage.Mongo
-import com.actionml.core.template.Engine
+import com.actionml.core.template.{Engine, RequiredEngineParams}
```
pferrel, line 6: RequiredParams moved to core/template/Engine.scala

rest-server/admin/src/main/scala/com/actionml/admin/RequiredEngineParams.scala
```diff
@@ -1,8 +0,0 @@
-package com.actionml.admin
-
-import com.actionml.core.template.EngineParams
-
-case class RequiredEngineParams(
-  engineId: String, // required, resourceId for engine
-  engineFactory: String
-) extends EngineParams
```
pferrel, line 8: moved to Engine

rest-server/build.sbt
```diff
@@ -21,10 +21,27 @@ lazy val commonSettings = Seq(
...
+    "org.apache.hadoop" % "hadoop-client" % "2.8.0",
+
+    "io.circe" %% "circe-core" % circeVersion,
+    "io.circe" %% "circe-generic" % circeVersion,
+    "io.circe" %% "circe-parser" % circeVersion,
     "org.typelevel" %% "cats" % "0.9.0",
 
     "com.typesafe" % "config" % "1.3.1",
     "com.iheart" %% "ficus" % "1.4.0",
+    "org.scaldi" %% "scaldi-akka" % "0.5.8",
```
pferrel, line 44: are we all using scaldi? @laser13 ?

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
```diff
@@ -18,31 +18,66 @@
...
   */
-abstract class Engine extends LazyLogging {
+abstract class Engine(implicit inj: Injector) extends LazyLogging with Injectable with JsonParser {
 
   // Todo: not sure how to require a val dataset: Dataset, which takes a type of Event parameter Dataset[CBEvent]
   // for instance. Because each Dataset may have a different parameter type
-  var engineId: String = _
+  val params: EngineParams
+  val algo: Algorithm
+  val engineId: String
```
pferrel: this is initialized in `initAndGet` so needs to be a var

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
```diff
@@ -18,31 +18,66 @@
...
-  def input(json: String, trainNow: Boolean = true): Validated[ValidateError, Boolean]
+  def input(json: String, trainNow: Boolean = true): Validated[ValidateError, Boolean] = {
+    mirroring.mirrorJson(engineId, json)
+    inputInternal(json: String, trainNow)
+  }
   def query(json: String): Validated[ValidateError, String]
   def status(): String = "Does not support status message."
+
+  protected def inputInternal(json: String, trainNow: Boolean = true): Validated[ValidateError, Boolean]
 }
```
pferrel: is `inputInternal` really needed? I'd rather remove it

rest-server/drivers/test_resource_2.json
```diff
@@ -1,6 +1,7 @@
 {
   "engineId": "test_resource_2",
   "engineFactory": "com.actionml.templates.cb.CBEngine",
+  "mirrorType": "hdfs",
```
pferrel, line 4: leaves the `mirrorLocation` out so should use the default.

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBEngine.scala
```diff
@@ -21,22 +21,29 @@ import cats.data.Validated
...
 
   override def init(json: String): Validated[ValidateError, Boolean] = {
-    parseAndValidate[CBEngineParams](json).andThen { p =>
-      params = p
-      engineId = params.engineId
-      dataset = new CBDataset(engineId)
-      algo = new CBAlgorithm(dataset)
-      Valid(p)
-    }.andThen(_ => algo.init(json, engineId))
+    super.init(json).andThen { _ =>
```
pferrel, line 39: `super.init` is how we pass the json to `Engine` for mirroring setup

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBEngine.scala
```diff
@@ -21,22 +21,29 @@ import cats.data.Validated
...
+class CBEngine(implicit inj: Injector) extends Engine() with JsonParser {
 
   var dataset: CBDataset = _
   var algo: CBAlgorithm = _
   var params: CBEngineParams = _
 
   override def init(json: String): Validated[ValidateError, Boolean] = {
-    parseAndValidate[CBEngineParams](json).andThen { p =>
-      params = p
-      engineId = params.engineId
```
pferrel, line 35: `engineId` is now set in `super.init` but there is no "contract" that requires an `algo` so it should be in `CBEngine` instead of `Engine`, same for `dataset`

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBEngine.scala
```diff
@@ -68,7 +75,8 @@ class CBEngine() extends Engine() with JsonParser {
   }
 
   /** Triggers parse, validation, and persistence of event encoded in the json */
-  def input(json: String, trainNow: Boolean = true): Validated[ValidateError, Boolean] = {
+  // Slava this needs to stay "input" why did you change the API?
+  def inputInternal(json: String, trainNow: Boolean = true): Validated[ValidateError, Boolean] = {
```
pferrel: Why do we need this new API?

snovgorodtsev-zz: I didn't channge the API, please see `com.actionml.core.template.Engine#input`

pferrel: I merged your personal develop branch. @snovgorodtsev, please push correction to `actionml/harness/mirror`

snovgorodtsev-zz: I just meant that `inputInternal` is not a part of public API. I am ok with removing this method but I do not know how to implement JSON mirroring on `Engine` level another way.

pferrel: why not call it `input`? Then have it call `super.input` something like below. That way the implementation can be put in `Engine` for all Templates to inherit.

```
super.input(json).andThen { p =>
  logger.trace("Got JSON body: " + json)
  // validation happens as the input goes to the dataset
  dataset.input(json).andThen(process(_)).map(_ => true)
}
```

PR #17, opened 2017-06-19
removing eventId from all events, they are not required but we need t…

pferrel:
…o remove the assumption from the routers etc

@laser13 the `eventId` cannot be assumed to exist for all Templates/engines. So it should be removed from routes, the individual event is not addressable in "Kappa" learners, which discard events after use so any route with an `eventId` is invalid at present.

`GET /engines/resource-id/events/eventId` is invalid as well as any other verb like POST, etc.

See code here: https://github.com/actionml/harness/blob/hotfix-eventId/rest-server/server/src/main/scala/com/actionml/router/http/routes/EventsRouter.scala#L35

other references in this file too. I'm not sure this router is needed. Also: https://github.com/actionml/harness/blob/hotfix-eventId/rest-server/server/src/main/scala/com/actionml/router/service/EventService.scala#L25


pferrel:
@laser13 I can try to remove these if you can give me a pointer? Basically 

POST `engines/resource-id/events` is valid

ANY VERB `engine/resource-id/events/eventId` is invalid

pferrel:
I took the `eventId` out of required event JSON fields and events with no IDs now works fine. The routes should be removed but unless someone uses them they will cause no error.

This should be the lowest priority work, leave it till last.

PR #19, opened 2017-07-07
Mirror inject

pferrel:
adding injection of mirroring object


rest-server/admin/src/main/scala/com/actionml/admin/Administrator.scala
```diff
@@ -40,7 +40,7 @@ abstract class Administrator extends LazyLogging {
     engineJson: Option[String] = None,
     dataDelete: Boolean = false,
     force: Boolean = false,
-    input: Option[String] = None): Validated[ValidateError, Boolean]
+    input: Option[String] = None): Validated[ValidateError, Unit]
```
pferrel: This is a core API we don't change these without agreed reasons because all templates must provide them. Generally the Server is more flexible than the Template APIs which I will be resistant to changing because they have to do with Machine learning workflow. I do change them but only to fit a requirement of our example Template, the Contextual Bandit.

We will eventually have several Templates that are separately compiled and are expected to implement the core Template APIs, if we change anything there all Templates will break and have to be modified at the API level. These will eventually be annotated so we don't change them and Template developers can rely on them.

So why pass no information back on Valid input? We convert sometimes from internal `Validated[ValidateError, SomeCaseClass]` to `Validated[ValidateError, Boolean]` when the case class has no further use but Unit?

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -72,15 +71,15 @@ class MongoAdministrator extends Administrator with JsonParser with Mongo {
...
-          logger.trace(s"Re-initializing engine for resource-id: ${params.engineId} with new params $json")
+          logger.trace(s"Re-initializing engine for resource-id: ${ params.engineId } with new params $json")
           val query = MongoDBObject("engineId" -> params.engineId)
           val update = MongoDBObject("$set" -> MongoDBObject("engineFactory" -> params.engineFactory, "params" -> json))
           enginesCollection.findAndModify(query, update)
         } else {
           //add new
-          logger.trace(s"Initializing new engine for resource-id: ${params.engineId} with params $json")
+          logger.trace(s"Initializing new engine for resource-id: ${ params.engineId } with params $json")
           val builder = MongoDBObject.newBuilder
```
pferrel: Let's not fix styling in PRs, it just makes them overly hard to read and we haven't agreed on style guidelines.

rest-server/core/src/main/scala/com/actionml/core/backup/FSMirroring.scala
```diff
@@ -18,70 +18,67 @@
...
+          mirrorEventError(s"$errMsg: ${ex.getMessage}")
       }
-
-    } else {
-      logger.warn("Local filesystem mirroring called, but not configured.")
-      noError = false
-    }
-    noError
+    Valid(())
   }
```
pferrel: We have been using Validated only when input is parsed or external commands are sent in, not for general error return values. Is this used later building the response to input?

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
```diff
@@ -20,43 +20,45 @@ package com.actionml.core.template
...
+    def createMirror(p: RequiredEngineParams) = {
+      p.mirrorContainer.fold[Validated[ValidateError, Unit]](Invalid(MissingParams("mirror container is undefined"))) {
+        container =>
+          p.mirrorType.fold[Validated[ValidateError, Unit]](Invalid(MissingParams("mirror type is undefined"))) {
+            case "fs" => mirroring = new FSMirroring(container); Valid(())
+            case mt   => Invalid(WrongParams(s"mirror type $mt is not implemented"))
+          }
+      }
+    }
+
```
pferrel: This `def` seem like a lot of stuff defined inline that should be moved out of the `init` method, otherwise it is hard to read a simple thing.

```
def init(json: String): Validated[ValidateError, Unit] = {
  parseAndValidate[RequiredEngineParams](json).andThen {
    createMirror
  }
}
```

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBEngine.scala
```diff
@@ -30,7 +30,7 @@ class CBEngine() extends Engine() with JsonParser {
   var algo: CBAlgorithm = _
   var params: CBEngineParams = _
 
-  override def init(json: String): Validated[ValidateError, Boolean] = {
+  override def init(json: String): Validated[ValidateError, Unit] = {
```
pferrel: This is not a big deal but for the core API I would not make such global changes without consulting first. They need to be stabilized as quickly as possible. I change them because of an algorithm or machine learning workflow need or some CLI need so ask about these first. Anything in core.templates.

rest-server/templates/src/main/scala/com/actionml/templates/cb/CBEngine.scala
```diff
@@ -142,8 +142,8 @@ Query
 }
 */
 case class CBQuery(
-    user: String,
-    groupId: String)
```
pferrel: http://docs.scala-lang.org/style/declarations

double indentation of vertically justified parameter lists in class defs

In general do not change my style in my code and I will do the same for you unless it violates the Scala style guide. If it's important we should agree but generally if the style is allowed in the Scala style docs we allow the author and maintainer to use what they want in code they maintain.

pferrel: I mention it here because in this case I believe the change is wrong, no?

pferrel:
ok so injection is the next step?

snovgorodtsev-zz:
Do you mean `scaldi` injection? I am just not sure if it's needed here....

pferrel:
We need each engine to be able to turn on and off mirroring based on config in their engine.json file. Injection would be nice when we have more than one way to mirror but not needed yet.

I actually like the idea of `Valid(())` instead of using `boolean` I was just pointing out that any API in `core.templates` must become stable soon and should be discussed if a change is needed.

If per engine mirroring is done and tested, go ahead and merge with develop. Make sure to take out the global mirroring config I did, just for cleanup.

pferrel:
I beleive we use `scaldi` elsewhere in the server. You should check.

snovgorodtsev-zz:
Ok I've cleaned up the code and the config a bit, please check it out. I also can move back the `Valid(())` if the team agree... And honestly speaking I still can not understand how `scaldi` could be applied here.

pferrel:
Do you mean injection or the `scaldi` lib? Use or not of `scaldi` is your choice. 

Even without injection we can have one type of mirroring per engine and it can be turned on/off with the `harness update -c engine.json` command. Injection only adds the ability to change the type of mirroring while the Harness Server is running. And since we are ok with localfs only now, all is well.

pferrel:
I think `Valid(())` or `Validated[ValidationError, Unit]` is a good idea to replace `Boolean` and would be ok to add to core. What do you think @laser13 ?

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -133,16 +132,10 @@ class MongoAdministrator extends Administrator with JsonParser with Mongo {
         engine.destroy()
         engine.init(params)
       }
-      if(input.nonEmpty) {
-        if( engines(engineId).importEvents(engines(engineId), input.get) ) Valid(true) else
-          Invalid(ValidRequestExecutionError(s"Unable to import from: ${input.get} on the servers file system to engineId:" +
-            s" ${engineId}. See server logs"))
```
pferrel, line 139: Is this error not possible?

rest-server/core/src/main/scala/com/actionml/core/backup/FSMirroring.scala
```diff
@@ -18,70 +18,60 @@
...
   val f = new File(mirrorContainer)
-  if( f.exists() && f.isDirectory() && config.getString("mirror.type").nonEmpty) {
-    logger.info(s"Morroring raw un-validated events to ${mirrorContainer}")
-  } else if(config.getString("mirror.type").nonEmpty){
-    logger.error(s"Mirror location: ${mirrorContainer} not configured to accept event mirroring.")
-  } else {
-    logger.warn(s"Mirroring location is set but type is not configured, no mirroring with be done.")
-  }
-
-
```
pferrel, line 44: Are you saying that these error conditions can never be seen?

rest-server/core/src/main/scala/com/actionml/core/backup/FSMirroring.scala
```diff
@@ -18,70 +18,60 @@
...
 import scala.io.Source
 
 /**
   * Mirroring implementation for local FS.
   */
 
-// TODO: not using injection so need a trait
-//object FSMirroring extends Mirroring {
-trait FSMirroring extends Mirroring {
+class FSMirroring(mirrorContainer: String) extends Mirroring(mirrorContainer) {
```
pferrel, line 34: I don't believe we can set the `mirrorContainer` at construction time since it is configured the `engine.json` file which is subject to update while the server is running.

rest-server/core/src/main/scala/com/actionml/core/backup/FSMirroring.scala
pferrel: I agree these should not be taken from the env using akka `ConfigFactory` but they still come from `engine.json` which can be updated while the server runs. So I put initialization in `init` since it is called when the Server's REST `PUT engines/resource-id/engines` is invoked. This updates mirror config by potentially passing in a new `engine.json`, which should have these params in it.

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
```diff
@@ -20,43 +20,40 @@ package com.actionml.core.template
...
+  private def createMirror(params: RequiredEngineParams) = params.mirrorContainer
+    .fold[Validated[ValidateError, Boolean]](Invalid(MissingParams("mirror container is undefined")))(
+    container => params.mirrorType
+      .fold[Validated[ValidateError, Boolean]](Invalid(MissingParams("mirror type is undefined"))) {
+      case "fs" => mirroring = new FSMirroring(container); Valid(true)
+      case mt   => Invalid(WrongParams(s"mirror type $mt is not implemented"))
+    })
+
+  def init(json: String): Validated[ValidateError, Boolean] = parseAndValidate[RequiredEngineParams](json)
+    .andThen(createMirror)
```
pferrel, line 47: Ah, ok I see how you are creating a new mirror class during runtime 👍

rest-server/server/src/main/resources/reference.conf
```diff
@@ -9,12 +9,4 @@ rest-server {
...
-}
-
-
-mirror {
-  type = "localfs"
-  type = ${?MIRROR_TYPE}
-  container = ""
-  container = ${?MIRROR_CONTAINER_NAME}
-}
+}
```
pferrel: 👍

pferrel:
Ok I understand what you are doing to support runtime changing of mirror so go ahead and merge. My comments may be incorrect now that I understand how you create the mirror in `engine.init`

snovgorodtsev-zz:
shell I merge it to `develop` now?

pferrel:
yes, if you have tested multiple engines with different mirroring setups in their engine.json files
Thanks

PR #20, opened 2017-08-02
Feature/user authentication

qqmbr4k:
Work in progress. Just sketch of directives and local auth module


pferrel:
We need to use Scala 2.11 there are libs we require that are not on 2.12 yet

andreyshikov:
closed - wip

PR #21, opened 2017-08-04
Feature/user authentication

qqmbr4k:


rest-server/server/src/main/resources/reference.conf
```diff
@@ -10,3 +10,8 @@ rest-server {
   port = 9090
   ssl = false
 }
+
+auth {
+  enabled = false
```
pferrel, line 15: Cool.

When we create these compiled in setting we also have an environment variable to override them. So far all env is set or checked here:  https://github.com/actionml/harness/blob/master/rest-server/bin/harness-env

rest-server/server/src/main/scala/com/actionml/security/directives/AuthDirectives.scala
```diff
@@ -0,0 +1,60 @@
...
+    } else provide(None)
+  }
+
+  def authorizeUser(role: Role, resourceId: ResourceId)
+                   (implicit as: ActorSystem,
+                    mat: ActorMaterializer,
+                    ec: ExecutionContext,
+                    secretOpt: Option[Secret],
+                    log: LoggingAdapter): Directive0 = {
+    if (config.auth.enabled) {
```
pferrel: We aren't that picky about style except to follow the scala style docs here: http://docs.scala-lang.org/, which I believe would have these params indented one indent. This is Java style.

rest-server/server/src/main/scala/com/actionml/security/model/User.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.actionml.security.model
+
+
+case class User(credentials: Secret,
+                role: Role,
+                resourceId: ResourceId)
```
pferrel: Scala style is 2 indents for class params.

pferrel:
Looks good so far.

pferrel:
@andreyshikov I will update the drawing here: https://docs.google.com/drawings/d/1_uPiP5UGkphF62PsdIM0Zu8Ugbp4XAKnLYP5CyKrfLo/edit

To include the offline authorization grant where we give toikens to users. Just to cover all the OAuth2 control flow. Correct me if I have it wrong.

rest-server/auth-server/src/main/scala/com/actionml/authserver/routes/AuthorizationController.scala
```diff
@@ -3,7 +3,7 @@ package com.actionml.authserver.routes
 import akka.http.scaladsl.model.headers.{BasicHttpCredentials, HttpChallenges}
 import akka.http.scaladsl.server.AuthenticationFailedRejection.{CredentialsMissing, CredentialsRejected}
 import akka.http.scaladsl.server._
-import com.actionml.authserver.model.Client
+import com.actionml.authserver.model.{Client, ClientId}
```
pferrel: I've taken to calling this "Users" in the docs.

andreyshikov: It's not user, it's client - the user's app server

pferrel: The word "client" means too many things. You are talking about an ID for something like a user account, right? Not a client to some server.

andreyshikov: No, it's about app server - same as at oauth2 spec

pferrel: Why did you want to call these "Users" I put that into the REST API URIs POST /users/...

pferrel: Can you check the new full REST API spec to make sure it is correct? Here: https://github.com/actionml/harness/blob/master/rest_spec.md

Note the use of `/users/<user-id>` in the Harness Rest-server API. Is the user-id what you are calling the `ClientId`?

pferrel: I'd rather leave the REST URIs referencing User collections, it doesn't have to match what you call it in the code neccessarily

PR #22, opened 2017-08-17
Update rest_spec.md

qqmbr4k:


rest_spec.md
```diff
@@ -63,15 +63,15 @@ These APIs allow the admin user to create new users with access to certain resou
...
 
 | HTTP Verb | URL | Request Body | Response Code | Response Body | Function |
 | --- | --- | :---  | :---  | :---  | :--- |
-| POST | `/authenticate` | `{"token": "string"}` | 200 or 401 | `{"accessToken": "string", "ttl": <optional long>, "refreshToken": "optional string"}` | authenticates user's access and returns a session token |
-| POST | `/authorize` | `{"accessToken": "string"`, ` "role": "string"`, ` "resourceId": "string"}` | 200 or 403 | `{"success": "true"}` | Given a session/access token authorize the access requested or return an error code |
+| POST | `/auth/token` | `grant_type=password&username=user-id&password=granted-token`, also app server's credentials should be provied by Authorization Basic header (see [https://tools.ietf.org/html/rfc6749#section-4.3] for details) | 200 or 401 | `{"access_token": "string", "token_type": "", "refresh_oken": "optional string"}` | authenticates user's access and returns a session token |
```
pferrel, line 76: shouldn't this be `POST /token` since we removed the `/auth` from URIs

I've incorporated these, will update develop with my next push

rest_spec.md
```diff
@@ -63,15 +63,15 @@ These APIs allow the admin user to create new users with access to certain resou
 
 | HTTP Verb | URL | Request Body | Response Code | Response Body | Function |
 | --- | --- | :---  | :---  | :---  | :--- |
-| POST | `/users` | `{"roleSetId": "client\|admin","resourceId": "*\|some id"}` | See Collection responses | `{“bearerToken”: "token"}` | Create a new user and assign a bearer token, setup internal management of the user-id that does not require saving the bearer token and attached the named `roleSet` for the `resource-id` to the new user |
-| DELETE | `/users/user-id` | `{"roleSetId": "client\|admin", "resourceId": "*\|<some-user-id>"}` | See Item responses | `{"bearerToken": "token"}` | Removes the user and all access rules from the system |
+| POST | `/users/user-id/permissions` | `{"userId": "user-id", "roleSetId": "client\|admin","resourceId": "*\|some id"}` | See Collection responses | `{"userId": "user_id", “bearerToken”: "token"}` | Create a new user and assign a bearer token and user-id, setup internal management of the user-id that does not require saving the bearer token and attached the named `roleSet` for the `resource-id` to the new user |
```
pferrel: If we are creating a user-id then this should be `POST /users` since the user-id is not known until it is created or we need both `/users` to return the user-id and `/users/user-id/permissions` to create permissions for the user-id. 

I assume permissions is a collection of roleSets keyed to resourceId? So a user-id can write to several engine-ids by setting several permissions on the same user-id?

How are we creating the user-id/client-id? It should be the server creating the id, right?

andreyshikov: We should separate user creation and permission grant. User creation method `POST /users` can also be able to grant permissions with additional params.
I will add `POST /users` in another PR.

pferrel: I agree. `POST /users` can set some initial state of the new user's permission and return the user-id. Then creating or deleting permissions will be sufficient but we will also need `DELETE /users/<user-id>` I can make the changes without a PR and you can review.

pferrel: We also need `GET /users` to list all user-ids and other info. I'll do any of these that don't directly relate to Auth-Server required APIs.

PR #23, opened 2017-08-28
[WIP] Python sdk with auth

qqmbr4k:
Very straightforward implementation of python sdk authorization.

### DIFF
--- a/python-sdk/actionml/__init__.py
+++ b/python-sdk/actionml/__init__.py
@@ -457,3 +457,31 @@ class CommandClient(BaseClient):
 
     def cancel_command(self, command_id):
         return self.async_cancel_command(command_id).get_response()
+
+
+class UsersClient(BaseClient):
+
+    def __init__(self, url="http://localhost:9090", threads=1, qsize=0, timeout=5):
+        self.path = "/auth/users"
+        super().__init__(url, threads, qsize, timeout)
+
+    def create_user(self, user_id=None, user_secret=None, role_set_id=None, resource_id=None):
+        request = AsyncRequest("POST", self.path, user_id=user_id, user_secret=user_secret, roleSetId=role_set_id, resourceId=resource_id)
+        request.set_response_handler(self._ok_response_handler)
+        self._connection.make_request(request)
+        return request
+
+
+class PermissionsClient(BaseClient):
+
+    def __init__(self, user_perm_id, url="http://localhost:9090", threads=1, qsize=0, timeout=5):
+        self.path = "/auth/users/{}/permissions".format(user_perm_id)
+        super().__init__(url, threads, qsize, timeout)
+
+    def grant_permission(self, user_id=None, user_secret=None, user_perm_id=None, role_set_id=None, resource_id=None):
+        request = AsyncRequest("POST", self.path, user_id=user_id, user_secret=user_secret,
+                               roleSetId=role_set_id, resourceId=resource_id, userPermId=user_perm_id)
+        request.set_response_handler(self._ok_response_handler)
+        self._connection.make_request(request)
+        return request
+
--- a/python-sdk/actionml/__init__.py
+++ b/python-sdk/actionml/__init__.py
@@ -108,7 +108,9 @@ class BaseClient(object):
             threads=self.threads,
             qsize=self.qsize,
             https=self.https,
-            timeout=self.timeout
+            timeout=self.timeout,
+            client_id="Aladdin",
+            client_secret="OpenSesame"
         )
 
     def close(self):
@@ -192,10 +194,13 @@ class EventClient(BaseClient):
 
     def async_create(self, event, entity_type, entity_id,
                      target_entity_type=None, target_entity_id=None, properties=None,
-                     event_id=None, event_time=None, creation_time=None):
+                     event_id=None, event_time=None, creation_time=None,
+                     user_id=None, user_secret=None):
         """
         Asynchronously create an event.
-        :param event_id: 
+        :param user_id: end-user id
+        :param user_secret: end-user secret (bearer token)
+        :param event_id:
         :param event: event name. type str.
         :param entity_type: entity type. It is the namespace of the entityId and
           analogous to the table name of a relational database. The entityId must be
@@ -234,18 +239,20 @@ class EventClient(BaseClient):
 
         data["creationTime"] = time_to_string_if_valid(creation_time)
 
-        request = AsyncRequest("POST", self.path, **data)
+        request = AsyncRequest("POST", self.path, user_id=user_id, user_secret=user_secret, **data)
         request.set_response_handler(self._create_response_handler)
         self._connection.make_request(request)
         return request
 
     def create(self, event, entity_type, entity_id,
                target_entity_type=None, target_entity_id=None, properties=None,
-               event_id=None, event_time=None, creation_time=None):
+               event_id=None, event_time=None, creation_time=None,
+               user_id=None, user_secret=None):
         """Synchronously (blocking) create an event."""
         return self.async_create(event, entity_type, entity_id,
                                  target_entity_type, target_entity_id, properties,
-                                 event_id, event_time, creation_time).get_response()
+                                 event_id, event_time, creation_time,
+                                 user_id, user_secret).get_response()
 
     def async_get(self, event_id):
         """
--- a/python-sdk/actionml/__init__.py
+++ b/python-sdk/actionml/__init__.py
@@ -83,7 +83,7 @@ def time_to_string_if_valid(t):
 
 
 class BaseClient(object):
-    def __init__(self, url, threads=1, qsize=0, timeout=5):
+    def __init__(self, url, threads=1, qsize=0, timeout=5, user_id=None, user_secret=None):
         """Constructor of Client object."""
         self.threads = threads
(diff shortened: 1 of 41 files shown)


pferrel:
This looks like it hard-codes a client-id and token. 

How do we make this easy to use? Maybe use of the env for client-id and token. the Token should not be in env but it's location can be then the location is protected with permissions like ssh keys, the client-id can go into the env, right? The user admin or SDK user should never have to deal with the ids unless they are using the users or permissions CLI where they are managing other users

we should talk about this

pferrel:
ok for a first.  Are you building Vowpal Wabbit?

python-sdk/actionml/__init__.py
```diff
@@ -109,8 +109,8 @@ def __init__(self, url, threads=1, qsize=0, timeout=5):
             qsize=self.qsize,
             https=self.https,
             timeout=self.timeout,
-            client_id="Aladdin",
-            client_secret="OpenSesame"
```
pferrel: Oh, I liked Aladdin and OpenSesame. Where did you pick that up?

andreyshikov: wikipedia, basic auth example )

python-sdk/actionml/connection.py
```diff
@@ -203,6 +206,7 @@ def set_request(self, request):
 
 class ActionMLHttpConnection(object):
```
pferrel, line 207: the project name is Harness, not ActionML. The package is misnamed and should be `harness`

python-sdk/actionml/__init__.py
```diff
@@ -398,9 +400,9 @@ def send_query(self, data):
 
 
 class CommandClient(BaseClient):
```
pferrel, line 402: should be `CommandsClient` plural, we should change all java and python "client" classes to follow this naming since they deal with resource types, which are collections.

python-sdk/actionml/connection.py
```diff
@@ -34,6 +35,7 @@
 # logger
 logger = None
 DEBUG_LOG = False
+AUTH_ENABLED = False
```
pferrel, line 39: How is auth enabled?

python-sdk/examples/cb_integration_test_2.py
```diff
@@ -115,12 +117,17 @@ def execute_queries(client, file):
...
+        # time.sleep(0.5)
 
         print("Result: \"\nuser\": " + data[0] + response.json_body)
         count += 1
 
+
 if __name__ == '__main__':
+
+    user_id = '8c24137d-24ed-498d-a5d6-f21051db9ce5'
+    user_secret = 'sILCF14ljZZdnqbzzg9SDJemz6wDY3YMH1rdbZ4r7RLzsA2vIh5R9H68WSMJN6ry'
```
pferrel, line 129: these are created by the server, how can you hard code them? You must be assuming some previous session and these are already in the DB? A clean example would take these are parameters on the CLI.

rest-server/auth-server/src/main/resources/application.conf
```diff
@@ -9,10 +9,39 @@ auth-server {
...
+  }]
+
+  role-sets = [
+    {
+      id = "admin"
+      roles = [
+        "user_create",
+        "user_permissions",
+        "engine_create"
+      ]
```
pferrel, line 27: and admin can do anything to any resource, this doesn't seem to capture that. For instance the admin cannot do what a client can do.

rest-server/auth-server/src/main/resources/application.conf
```diff
@@ -9,10 +9,39 @@ auth-server {
...
+        "event_read",
+        "event_modify",
+        "query_create",
+        "query_modify",
+        "query_read"
+      ]
+    }
+  ]
+
   access-token-ttl = 3600000 // 1 hour
```
pferrel, line 43: Is an hour necessary? Could this be more like a day? Input can come in very fast, in bursts, and every time a new access token is required it will slow down the request. The access token is made to speed up auth so the credentials do not need to be checked on every request, right?

rest-server/auth-server/src/main/scala/com/actionml/authserver/dal/mongo/ClientsDaoImpl.scala
```diff
@@ -27,11 +27,12 @@ import scala.concurrent.{ExecutionContext, Future}
...
+  private lazy val clients = collection[Client]("clients")
 
-  override def find(id: String): Future[Client] = {
+  override def find(id: String): Future[Option[Client]] = {
     clients.find(equal("id", id))
       .first
       .toFuture
+      .map(Option.apply)
   }
 }
```
pferrel, line 38: I like the DAO pattern you are using. I use Salat for generic DAOs but they cause Mongo queries to be used everywhere. This is much cleaner.

rest-server/auth-server/src/main/resources/application.conf
pferrel: strictly speaking the only REST defined is:
```    
"engine_read",
"event_create",
"query_create",
```

Since events and queries are not persisted they cannot be read or modified.

rest-server/auth-server/src/main/scala/com/actionml/authserver/routes/AuthenticationDirectives.scala
```diff
@@ -25,11 +25,11 @@ import com.actionml.authserver.Realms
 import scala.concurrent.Future
 
 
-trait ClientAuthentication extends Directives {
+trait AuthenticationDirectives extends Directives {
 
-  def authenticateClient(fn: (String, String) => Future[_]): Directive1[String] = extractCredentials.flatMap {
-    case Some(BasicHttpCredentials(clientId, password)) =>
-      onSuccess(fn(clientId, password)).map(_ => clientId)
+  def basicAuth(fn: (String, String) => Future[_]): Directive1[String] = extractCredentials.flatMap {
```
pferrel, line 30: What do you mean by `basicAuth`? Are we using OAuth2 or not? If we are using OAuth2 what would you call the type, there are several types described in the OAuth2 spec.

rest-server/server/src/main/resources/reference.conf
```diff
@@ -13,5 +13,7 @@ rest-server {
 
 auth {
   enabled = false
-  auth-server-url = "http://localhost:9099/auth"
+  server-url = "http://localhost:9099"
+  client-id = "harness"
+  client-secret = "mwjmo4fwC9qFW081gm5vwsmAC4Z2747XEU8eeV6dzUzjW"
```
pferrel, line 18: these need cleanup or documentation for how to change in `harness-env`

rest-server/server/src/main/scala/com/actionml/router/http/routes/QueriesRouter.scala
```diff
@@ -40,7 +41,7 @@ class QueriesRouter(implicit inj: Injector) extends BaseRouter with AuthDirectiv
     }
   }
 
-  private def getPrediction(engineId: String, log: LoggingAdapter): Route = (get & asJson) { query =>
+  private def getPrediction(engineId: String, log: LoggingAdapter): Route = (post & asJson) { query =>
```
pferrel, line 44: Odd, how did this ever work as a `GET`???

rest-server/auth-server/src/main/resources/application.conf
andreyshikov: So, engine_create shouldn't be at any roleset? Should this role be granted manually later?

rest-server/auth-server/src/main/resources/application.conf
andreyshikov: > Is an hour necessary? Could this be more like a day?

It's not my magic number, it's recommendations of oauth2 spec https://tools.ietf.org/html/rfc6750#section-5.3

rest-server/auth-server/src/main/scala/com/actionml/authserver/routes/AuthenticationDirectives.scala
andreyshikov: oauth2 uses basic auth at some stages, it's absolutely corresponds to spec

rest-server/server/src/main/scala/com/actionml/router/http/routes/QueriesRouter.scala
andreyshikov: It was my mistake from very first branch, just fixing it

rest-server/auth-server/src/main/resources/application.conf
andreyshikov: @pferrel now we managing roles with roleSets. What roleSet should contain engine_create role?
Looks like it's neither admin nor client, so we need another roleSet?

rest-server/auth-server/src/main/resources/application.conf
andreyshikov: @pferrel updated rolesets

rest-server/auth-server/src/main/resources/application.conf
andreyshikov: @pferrel changed rolesets, admin now contains all the roles

PR #24, opened 2017-08-28
engine id same as at cb_integration_test

qqmbr4k:


pferrel:
Good catch, this worked for me only because I had those other engine-ids already created. You should change the other json engine file too.

pferrel:
no, to use with `harness/python-sdk/integration-test.sh` they have the correct engine-id in the develop branch

PR #30, opened 2017-09-07
Added auth support to java sdk

qqmbr4k:


java-sdk/example/src/main/java/EventClientExample.java
```diff
@@ -45,7 +45,14 @@ public static void main(String[] args) {
 
         log.info("Args: {}, {}, {}, {}", engineId, fileName, serverHost, serverPort);
 
-        EventClient client = new EventClient(engineId, serverHost, serverPort);
+        Map<String, String> env = System.getenv();
+        Optional<String> optUsername = Optional.ofNullable(env.getOrDefault("HARNESS_USER_ID", null));
+        Optional<String> optPassword = Optional.ofNullable(env.getOrDefault("HARNESS_USER_SECRET", null));
+        Optional<PasswordAuthentication> optionalCreds = optUsername.flatMap(username -> optPassword.map( password ->
```
pferrel: Can we make this HARNESS_CLIENT_USER_ID and HARNESS_CLIENT_USER_SECRET? User may have an admin user and a client user for 2 different purposes, client for sending events, admin for CLI

andreyshikov: Ok, I will add it to update

PR #34, opened 2017-09-08
Removed ...protected from harness env

qqmbr4k:


rest-server/bin/common.py
```diff
@@ -22,11 +22,9 @@
 harness_host = os.getenv('REST_SERVER_HOST', 'localhost')
 harness_port = os.getenv('REST_SERVER_PORT', 9090)
 
-if os.getenv('HARNESS_AUTH_SERVER_PROTECTED') == 'true' and os.getenv('HARNESS_AUTH_ENABLED') == 'true':
+if os.getenv('HARNESS_AUTH_SERVER_PROTECTED') == 'true' or os.getenv('HARNESS_AUTH_ENABLED') == 'true':
     auth_enabled = True
     print('Auth enabled')
```
pferrel, line 27: This misses the error condition where they are not set the same, this is always an error. HARNESS_AUTH_SERVER_PROTECTED should never be visible to harness unless it is set globally in .profile or .bashrc etc.

PR #38, opened 2017-09-18
Scaffold Template

pferrel:
This is a Scaffolding or skeleton Template that can be used to start from in creating a new Template/Engine. It does nothing but is more than the most minimal Template since it implements methods that most Template will want rather than relying on the abstract class behavior. It is meant as a starting point for new Templates but can also be run for things like performance tests since it actually takes and validates input and responds to queries but does nothing more than minimal work.

### DIFF
--- a/rest-server/core/src/main/scala/com/actionml/core/template/Dataset.scala
+++ b/rest-server/core/src/main/scala/com/actionml/core/template/Dataset.scala
@@ -22,9 +22,9 @@ import com.actionml.core.storage.Store
 import com.actionml.core.validate.ValidateError
 import com.typesafe.scalalogging.LazyLogging
 
-abstract class Dataset[T](r: String) extends LazyLogging {
+abstract class Dataset[T](engineId: String) extends LazyLogging {
 
-  val resourceId: String = r
+  val resourceId: String = engineId
 
   def init(json: String): Validated[ValidateError, Boolean]
   def destroy(): Unit
@@ -36,5 +36,18 @@ abstract class Dataset[T](r: String) extends LazyLogging {
   def parseAndValidateInput(s: String): Validated[ValidateError, T]
 
 }
+
+
+// allows us to look at what kind of specialized event to create
+case class GenericEvent (
+  //eventId: String, // not used in Harness, but allowed for PIO compatibility
+  event: String,
+  entityType: String,
+  entityId: String,
+  targetEntityId: Option[String] = None,
+  properties: Option[Map[String, Any]] = None,
+  eventTime: String, // ISO8601 date
+  creationTime: String) // ISO8601 date
+  extends Event
 
 trait Event
--- a/rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
+++ b/rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
@@ -34,7 +34,7 @@ abstract class Engine extends LazyLogging with JsonParser {
   var engineId: String = _
   var mirroring: Mirroring = _
 
-  private def createMirror(params: RequiredEngineParams): Validated[ValidateError, Boolean] = {
+  private def createMirror(params: GenericEngineParams): Validated[ValidateError, Boolean] = {
     if (params.mirrorContainer.isEmpty) {
       logger.info("No mirrorContainer defined for this engine so no event mirroring will be done.")
       Valid(true)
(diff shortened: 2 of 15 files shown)


pferrel:
There is no need to review in detail. This contains some core refactoring of Template classes and will take the existing input events and respond to existing queries (for the CBEngine). Therefore we can use this as a basis for load and performance tests.

rest-server/core/src/main/scala/com/actionml/core/template/Dataset.scala
```diff
@@ -22,9 +22,9 @@ import com.actionml.core.storage.Store
 import com.actionml.core.validate.ValidateError
 import com.typesafe.scalalogging.LazyLogging
 
-abstract class Dataset[T](r: String) extends LazyLogging {
+abstract class Dataset[T](engineId: String) extends LazyLogging {
 
-  val resourceId: String = r
+  val resourceId: String = engineId
```
andreyshikov, line 27: this engineId alias looks confusing for me

rest-server/core/src/main/scala/com/actionml/core/template/Dataset.scala
```diff
@@ -37,4 +37,17 @@ abstract class Dataset[T](r: String) extends LazyLogging {
...
+
+// allows us to look at what kind of specialized event to create
+case class GenericEvent (
+  //eventId: String, // not used in Harness, but allowed for PIO compatibility
+  event: String,
+  entityType: String,
+  entityId: String,
+  targetEntityId: Option[String] = None,
+  properties: Option[Map[String, Any]] = None,
+  eventTime: String, // ISO8601 date
```
andreyshikov, line 49: maybe use DateTime for eventTime and add iso parser below?

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
```diff
@@ -50,7 +50,7 @@ abstract class Engine extends LazyLogging with JsonParser {
     }
   }
 
-  def init(json: String): Validated[ValidateError, Boolean] = parseAndValidate[RequiredEngineParams](json)
+  def init(json: String): Validated[ValidateError, Boolean] = parseAndValidate[GenericEngineParams](json)
```
andreyshikov, line 53: I don't like vars (engineId and mirroring) and 'init' methods, and I think we can get rid of this later with something like 'factory method' pattern:
```Engine { def engineId }
CBEngine(eid: String) extends Engine { override def engineId = eid }
object CBEngine.apply(json: String)
```
but i'm just mentioning todos for future

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
```diff
@@ -67,21 +67,30 @@ abstract class Engine extends LazyLogging with JsonParser {
...
   mirrorContainer: Option[String] = None) extends EngineParams
 
+/** Used only for illustration since queries have no required part */
+case class GenericQuery() extends Query {
+  def toJson = "{\"dummyQuery\": \"query\""
+}
+
+/** Used only for illustration since query results have no required part */
+case class GenericQueryResult() extends QueryResult{
+  def toJson = "{\"dummyQueryResult\": \"result\""
```
andreyshikov: I like """ for that

qqmbr4k (approved):

rest-server/core/src/main/scala/com/actionml/core/template/Dataset.scala
pferrel: not for a scaffold, no data is actually used

rest-server/core/src/main/scala/com/actionml/core/template/Dataset.scala
pferrel: In the beginning Events and Queries were tied to Datasets and Engines. But experience showed an Engine is needed in all cases as a controller of the Dataset, so what was called a resourceId "r" is now always an engineId and so this should be less confusing I hope.

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
pferrel: We use it so Validated can be returned from Init since this is the only place the JSON Engine params file is validated and parsed. Apply would not allow for this. But we can discuss.

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
pferrel: yeah, and also a final }
:-)

PR #39, opened 2017-09-22
Feature/harness auth cache

qqmbr4k:
Added cache for harness access tokens


rest-server/server/src/main/scala/com/actionml/authserver/services/CachedAuthorizationService.scala
```diff
@@ -0,0 +1,65 @@
...
+        tokenCache.put(key, result)
+        result
+      }
+    }
+  }
+}
+
+
+object CachedAuthorizationService {
+  val cacheTtl = Duration.of(30, TimeUnit.MINUTES)
```
pferrel, line 50: We should probably have all timeouts and limits in general defined in config with a method to override with harness-auth-env.sh or plain old env vars. This will clutter things but we can also choose to not add them to not put them in harness-auth-env.sh but they are there if we need them.

pferrel: I plan to deploy this Thursday so clean up anything you need to before then.

pferrel:
I'm trusting you, a little to complicated to follow by reading :-)

PR #41, opened 2017-10-23
ehcache workaround

qqmbr4k:


rest-server/server/src/main/scala/com/actionml/authserver/services/CachedAuthorizationService.scala
```diff
@@ -57,9 +57,9 @@ object CachedAuthorizationService {
   private val cacheManager: CacheManager = CacheManagerBuilder.newCacheManagerBuilder
     .withCache("access_tokens", CacheConfigurationBuilder.newCacheConfigurationBuilder(
       classOf[(AccessToken, RoleId, ResourceId)],
-      classOf[Boolean],
+      classOf[java.lang.Boolean],
```
pferrel, line 60: Very odd fix. Is the ecache library Java? Are you sure this is the right fix?

pferrel:
LGTM

PR #42, opened 2017-10-27
Bugfix/user roles

qqmbr4k:
Added strict compliance for config roleSetIds and users's roleSetIds. Previous implementation reckoned that if one role id set contains another then user has that role.


pferrel:
Is this PR to enforce that all users have only one roleSetId? 

Does this mean a user can have access as "client" to only one engine? If so it is not how it should be. A user should be able to access many engines if given "client" permission for them. In fact this is typical.

"client" user will have many engines that they can access
"admin" is allowed access to all routes, engines, events, queries. 

There can be many admins and clients. Clients can overlap in which engines they have access to so client A can access engine J, K, N and client B can access K, N, M This is not typical but should be allowed. More typically the access is the same for 2 client users client A can access engine J, K, N and client B can access J, K, N.

PR #43, opened 2017-12-28
Hinting works, time for review

pferrel:
Still have the ML model persistence to implement for Harness restart but functional tests are successful in a single Harness session.


rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -46,7 +46,17 @@ class MongoAdministrator extends Administrator with JsonParser with Mongo {
...
+      if (e._2 == null) { // it is possible that previously valid metadata is now bad, the Engine code must have changed
+        logger.error(s"Error creating engineId: $engineId from $params" +
+          s"\n\nTrying to recover by deleting the previous Engine metadata but data may still exist for this Engine, which you must " +
+          s"delete by hand from whatever DB the Engine uses then you can re-add a valid Engine JSON config and start over. Note:" +
+          s"this only happens when code for one version of the Engine has chosen to not be backwards compatible.")
+        // Todo: we need a way to cleanup in this situation
+        enginesCollection.remove(MongoDBObject("engineId" -> engineId))
+        // can't do this because the instance is null: deadEngine.destroy(), maybe we need a compan ion object with a cleanup function?
+      }
+      e
```
pferrel, line 59: This indicates a flaw in Harness logic about someone messing up the engine.json file with invalid setting. We need a much better way to report the error than this allows. Also need a better way to clean up something that was partially created but eventually the JSON was found to have an error. Different classes get a chance to validate and if valid do some init that may be persistent, once the error is found this init may need cleanup

pferrel:
should be against develop

PR #44, opened 2017-12-28
Hinting review

pferrel:
Still have the ML model persistence to implement for Harness restart but functional tests are successful in a single Harness session.


rest-server/templates/src/main/scala/com/actionml/templates/navhinting/NavHintingAlgorithm.scala
```diff
@@ -0,0 +1,219 @@
...
+  /** add the event to the end of an active journey subject to length limits */
+  def updateTrail(navId: String, timeStamp: DateTime, trail: Seq[(String, DateTime)]): Seq[(String, DateTime)] = {
+    val newTrail = trail :+ (navId, timeStamp)
+    newTrail.takeRight(params.numQueueEvents.getOrElse(50))
+  }
+
+
+  /** update the model with a Future for every input. This may cause Futures to accumulate */
+  def updateModel(convertedJourney: Journey,  now: DateTime): Unit = {
+    applyDecayFunction(convertedJourney, now).map { weightedVectors =>
```
pferrel, line 116: this maps a Future

pferrel: Do we have the possibility to have overlapping Futures here, both modifying the `var model`?

andreyshikov: summarizing slack conversation: we can combine these two futures and return resulting future to the upper level. e.g. like that:
`def updateModel(...): Future[Unit] = {
    applyDecayFunction.map( <merge operation> ).flatMap(<write model file>)`

pferrel: done, it will show up in the next PR

PR #45, opened 2018-01-09
Feature/security defaults

qqmbr4k:
Removed keys from the repository, disabled SSL by default, moved SSL config to separate file.


rest-server/auth-server/src/main/resources/application.conf
```diff
@@ -1,11 +1,17 @@
+include "akka.conf"
+include "akka-http.conf"
+
+//uncomment line below to include ssl configuration
+//include "akka-ssl.conf"
```
pferrel, line 5: Really, uncomment and re-compile? We want to avoid recompiling and be able to switch SSL on and off via config and restart, right?

pferrel:
Yeah too many file changes to see what the refactor does without a little documentation. There are docs in the base dir for SSL and Auth that you can change, mark your changes so I can edit if needed.

pferrel:
https://github.com/actionml/harness/blob/develop/security.md
and at the bottom of this file
https://github.com/actionml/harness/blob/develop/install.md

rest-server/auth-server/src/main/resources/application.conf
andreyshikov: > Really, uncomment and re-compile? 

no, we use the external config file (it's -Dresource.config=... param). Dist script uses this file as a template. So, no recompilation is needed.

pferrel: But to change the setting `-Dresource.config=...` means re-compiling, correct? 

Ideally we want just a **restart** with new config to change SSL or Auth.

andreyshikov: > But to change the setting -Dresource.config=... means re-compiling, correct?

No, it's a file system path.

> Ideally we want just a restart with new config to change SSL or Auth.

Yeah, we can make some changes to the config (somewhere on the file system) and restart.

install.md
```diff
@@ -153,13 +153,38 @@ This creates a file in the admin user's .ssh directory containing the secret. No
...
 ## Setup Server-side TLS/SSL 
 
-TLS must be enabled on the Harness Server and on all SDKs that communicate with it. That means at very least Python must be configured because the CLI uses the Python SDK.
+TLS can be toggled via HARNESS_SSL_ENABLED env variable (true|false). When set to 'true' it must be enabled on the Harness Server and on all SDKs that communicate with it. That means at very least Python must be configured because the CLI uses the Python SDK.
+
+    export HARNESS_SSL_ENABLED=${HARNESS_SSL_ENABLED:-false}
+TLS configuration can be configured at the file that HARNESS_SSL_CONFIG_PATH points to.
+
+    export HARNESS_SSL_CONFIG_PATH=${HARNESS_SSL_CONFIG_PATH:-conf/akka-ssl.conf}
+The content of the file is similar to the Akka SSL configuration:
```
pferrel: This path will never change why surface it as an env var? Also it will not exist in some future binary release. I really don't understand the need to look at .conf at runtime. The .conf should be set to override some things with env vars, but should never be references by env vars.

We really need to move away from config known only at compile time unless it does not change but we want to allow someone to change it and re-compile. All user changeable settings should not reference .conf files since in the future they will not be distributed in the binary dist.

PR #47, opened 2018-02-13
Harness update

pferrel:
Server seems to work but can't run integration test due to messed up SSL config


java-sdk/src/main/resources/akka.conf
```diff
@@ -15,31 +15,31 @@ akka {
...
+  ]
+
+  ssl-config.debug = {
+    ssl = true
+    handshake = true
+    verbose = true
+    data = true
+    trustmanager = true
+    keymanager = true
+  }
```
pferrel: I tried commenting this out and leaving it in for the sdk, but neither works with the server

nav-hinting.md
```diff
@@ -131,6 +131,7 @@ The NH Engine has a configuration file defined below. This defines parameters fo
   "engineId": "hinting",
   "engineFactory": "com.actionml.templates.nh.NavHintingEngine",
   "algorithm":{
+    "journeyTTL": 90, // number of days, 90 is default
```
pferrel, line 134: This is to set the TTL on the journey's collection, not used yet but the setting is available in the config params

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -106,6 +106,22 @@ class MongoAdministrator extends Administrator with JsonParser with Mongo {
...
+  override def updateEngine(json: String): Validated[ValidateError, String] = {
+    parseAndValidate[GenericEngineParams](json).andThen { params =>
+      val existingEngineOpt = engines.get(params.engineId)
+      if (existingEngineOpt.isDefined) { // found the engine to update
+        // re-initialize
+        logger.trace(s"Re-initializing engine for resource-id: ${params.engineId} with new params $json")
+        val query = MongoDBObject("engineId" -> params.engineId)
+        val update = MongoDBObject("$set" -> MongoDBObject("engineFactory" -> params.engineFactory, "params" -> json))
+        enginesCollection.findAndModify(query, update)
+        existingEngineOpt.get.init(json, deepInit = false).andThen(_ => Valid(json))
```
pferrel: This is the same as adding or instantiating and engine from persisted data **except** the initialization is not "deep". This is interpreted by the Engine, but usually means not do re-init the algorithm since it is running and you cannot alter params for an existing model, the engine needs to be deleted along with it's model and re-added to init from a clean starting point.

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -106,6 +106,22 @@ class MongoAdministrator extends Administrator with JsonParser with Mongo {
     }
   }
 
+  override def updateEngine(json: String): Validated[ValidateError, String] = {
+    parseAndValidate[GenericEngineParams](json).andThen { params =>
+      val existingEngineOpt = engines.get(params.engineId)
+      if (existingEngineOpt.isDefined) { // found the engine to update
```
andreyshikov: It also can be
```
engines.get(params.engineId).map { existingEngine =>
... }
.getOrElse(Invalid(WrongParams(s"Unable to update Engine: ${params.engineId}, the engine does not exist")))
```
less variables, no fooOpt.get, and bit more clear (not just style changes)

java-sdk/src/main/resources/akka.conf
andreyshikov: @pferrel  can I merge feature/security_defaults here?

qqmbr4k (changes requested):
Can I merge feature/security_defaults into this branch?

qqmbr4k (approved):

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
pferrel: True this can be done but is it clearer? I am conflicted about some Scala idioms that seem clearer to a Scala programmers but are much harder for a non-Scala programmers to understand, including map on an Option which doesn't return the thing contained but the identical Option. Not to mention .map code that is not executed if the Option == None. My code makes that quite explicit. I've been using Scala for some time now and find some of these things very non-intuitive. This causes me to resist some fully idiomatic uses of Scala, which as a language, is highly criticised and resisted by Java developers for things like this. So I remain conflicted. Java has most of these features now (Java 9) so I suppose you are correct and I should start accepting more idiomatic Scala. In my mind .map is overloaded to do too much magic, like create callbacks when used on Futures, or this when used on Options. I wish they would use some other word because .map becomes too hard to interpret unless you are very used to the idioms.

I do, however,  fully understand your comment and yes, it would be more clear to an experienced Scala dev.

pferrel: I will say that I 100% disagree with people who try to use as few lines of code to do something as possible. This non-logic leads to often unreadable code and even if theory says there will be less bugs (less lines of code), I would trade readability for bugs any day. A line of code should be modifiable and that requires being able to read and understand it easily. I know this is not what you are suggesting, just my opinion on a related point

andreyshikov: It's not about idioms. Option.get can throw NPE. Less variable declarations look better.

pferrel: `engines.get(params.engineId)` is getting a value from a Map, not a value from an Option. As I understood `someMap.get(key)` would create an Option == None if the key did not exist and Option(value) when the key does exist. This is to avoid `someMap(key)`, which does throw an exception when the key does not exist. I don't see where an exception can possibly be thrown. 

Later when I get from the Option, I have already checked that it is defined, so no exception there, right?

Did I miss some case?

So this is about idioms, and style and yes, I understand that point. But if you are saying there is a logic flaw that will allow an exception, that will require a change.

pferrel: Sorry to belabor this point, it seems important. I am not trying to join in a useless debate :-)

andreyshikov: No, it's ok, debate is quite important.
> Did I miss some case?
yes, I mean `existingEngineOpt.get.` at line 118

pferrel: but that is inside a block that checks to see if `existingEngineOpt.isDefined` so that cannot cause an exception, right?

pferrel: If this is not true, I use it in other places and they will all need to be checked.

andreyshikov: yes, the code works correctly (from my point of view), if (option.defined) option.get is like checking for null. If we use Scala, why don't use language features instead of trying to write go or java7-?

pferrel: yes, I understand the style/idiom point. I'll try rewriting it and see if it seems easier to read. But as I have said, heavy use of Scala idioms IMHO can obscure the meaning of the code in some cases.

I reserve the right to not use language features that I think hide meaning. For example when I started using C++ in it's very early days we had a rule to never use operator overloading unless the meaning was unquestionably obvious. Scala suffers from this too, different but related collections use different operators for essentially the same operation. There may be reasons for this and I do use operator overloading in Scala when creating a DSL. But anyway, just because a feature exists in a language does not mean it is good, It only means that the language designer disagrees with my definition of good.

andreyshikov:
No need to rewrite the code, just mentioned that those cases can be written different way in the future

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
pferrel: OK, the code seems readable if you know what the map returns, and that is deeply idiomatic.

```
      engines.get(params.engineId).map { existingEngine =>
        logger.trace(s"Re-initializing engine for resource-id: ${params.engineId} with new params $json")
        val query = MongoDBObject("engineId" -> params.engineId)
        val update = MongoDBObject("$set" -> MongoDBObject("engineFactory" -> params.engineFactory, "params" -> json))
        enginesCollection.findAndModify(query, update)
        existingEngine.init(json, deepInit = false).andThen(_ => Valid(
          """{
            |  "comment":"Get engine status to see what was changed."
            |}
          """.stripMargin))
      }.getOrElse(Invalid(WrongParams(s"Unable to update Engine: ${params.engineId}, the engine does not exist"))
```

In this case the .map returns an Option(Validated...) not a Valdated...  as one would expect. This is one thing about Scala that I dislike, it is very hidden and this is why people shy away from using it. But yes it is more idiomatic so I'll add it.

pferrel: I suspect that my understanding of `.map` is incomplete somehow.

pferrel: ok changed to the above.

pferrel:
Conflicts resolved. Will be pushed to develop

PR #48, opened 2018-02-19
Async mongo with develop

qqmbr4k:


install.md
```diff
@@ -153,12 +153,38 @@ This creates a file in the admin user's .ssh directory containing the secret. No
...
+          ]
+        }
+      }
+    }
+
+also path to the keystore, it's type and password can be set via
+
+    export HARNESS_KEYSTORE_PASSWORD=${HARNESS_KEYSTORE_PASSWORD:******}
+    export HARNESS_KEYSTORE_TYPE=<e.g. JKS(default) or PKCS12>
+    export HARNESS_KEYSTORE_PATH=${HARNESS_KEYSTORE_PATH:-$HARNESS_HOME/harness.jks}
```
pferrel, line 185: We should have a default password documented here and other defaults should be here unless the doc is about how to change these default values.

install.md
```diff
@@ -185,7 +211,7 @@ To use the examples in `java-sdk/examples` you can also set the following env:
 
 The following must be setup to use either the Java or Python SDK
 
-    export HARNESS_CA_CERT=/path/to/harness/server/cert.pem
+    export HARNESS_SERVER_CERT_PATH=/path/to/harness/server/cert.pem
     # points to the server's .pem file, used for self-signed
```
pferrel, line 215: As a convention I don't descript paths this way anymore. Non-english speakers find this confusing. So I do something like <!!!put-your-path-here!!!> or something that will cause an error if they to not replace it, instead of a "cannot find". I get too many questions about this on mailing lists for things like PIO and the UR.

java-sdk/example/data/bqtech-bug-query.json
```diff
@@ -0,0 +1,4 @@
+{"groupId":"10","user":"648b4f62-0f22-453c-98f2-f8e2bc04f5ae-10"}
+{"groupId":"10","user":"648b4f62-0f22-453c-98f2-f8e2bc04f5ae-10"}
+{"groupId":"10","user":"648b4f62-0f22-453c-98f2-f8e2bc04f5ae-10"}
+{"groupId":"10","user":"648b4f62-0f22-453c-98f2-f8e2bc04f5ae-10"}
```
pferrel, line 4: This file should not be part of the repo

java-sdk/nav-hinting-integration-test.sh
```diff
@@ -0,0 +1,80 @@
...
+host=localhost
+engine=hinting
+test_queries=data/nh-queries.json
+user1_events=data/nh-pferrel-events.json
+user2_events=data/nh-joe-events.json
+user1_conversion=data/nh-pferrel-conversion.json
+user2_conversion=data/nh-joe-conversion.json
+sleep_seconds=1
+
+export HARNESS_SERVER_CERT_PATH=/Users/pat/harness/rest-server/server/src/main/resources/keys/harness.pem
```
pferrel, line 18: Should this be in the script? I don't think so. The test should run with Harness as it is configured, requiring no config in the script.

java-sdk/pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.actionml</groupId>
     <artifactId>harness-java-sdk</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
+    <version>0.1.0</version>
```
pferrel, line 10: technically Harness and the SDKs are on 0.1.1-SNAPSHOT or -RC1

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -148,7 +162,7 @@ class MongoAdministrator(override implicit val injector: Module)
       // Todo: implement
       logger.info("Using 'harness update -c <some-engine-json-file>' is not implemented yet")
```
pferrel, line 163: -c is not required, this is an old error message

pferrel (approved):
Looks ok as far as I can tell.

install.md
andreyshikov: This one and other comments below are about fixes|improvements, but this PR is about the merge. I'll do those fixes in next PR.

PR #50, opened 2018-02-25
Async mongo wip

pferrel:
tracking move to async mongo client and DAL/DAO/DAOImpl pattern


rest-server/core/src/main/scala/com/actionml/core/dal/CBGroupsDao.scala
```diff
@@ -0,0 +1,31 @@
+/*
```
pferrel: This should be in the template package under the CB engine since it is Engine Specific. The Entire DAL/DAO/DAOImpl should be moved out of core. Once or if we find how to make these use Parameterized Types we may be able to move to core. But that is not required, I'd rather see if there is enough benefit from the extra work. We have no hard require to do yet another DB virtualization layer but it would make users job of creating Engines much easier if we could.

pferrel:
Is this supposed to be done?

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -17,53 +17,77 @@
...
+        .map(Success(_))
+        .recoverWith { case error =>
+          // it is possible that previously valid metadata is now bad, the Engine code must have changed
+          logger.error(s"Error creating engineId: ${engine._id} from ${engine.params}" +
+            s"\n\nTry deleting the engine instance by hand and re-adding to start fresh or fix the JSON config if there " +
+            s"is an error in it. Note: " +
+            s"this may happen when code for one version of the Engine has chosen to not be backwards compatible.")
+          // Todo: we need a way to cleanup in this situation
+          enginesCollection.deleteOne(Document("engineId" -> engine._id)).toFuture().map(_ => Failure(error))
+          // can't do this because the instance is null: deadEngine.destroy(), maybe we need a compan ion object with a cleanup function?
```
pferrel: don't delete it from the DB. I have changed the error message. Do nothing but report the error and return an error code to the REST API.

Also `${engine._id}` should be the engineId. You should re-define the case class you use for the EngineInstance data in Mongo to have `_id: String` and use the engineId as the _id key.

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -77,62 +101,43 @@ class MongoAdministrator extends Administrator with JsonParser with Mongo {
   Success/failure indicated in the HTTP return code
   Action: creates or modifies an existing engine
   */
-  def addEngine(json: String): Validated[ValidateError, EngineId] = {
+  def addEngine(json: String)(implicit ec: ExecutionContext): Future[Validated[ValidateError, EngineId]] = {
     // val params = parse(json).extract[GenericEngineParams]
```
pferrel, line 102: Is the ExecutionContext where we can use Scaldi to inject from? I'd like the inject the Mongo implementation for the DAOs so the Engine do not need to deal with Mongo specific details. Same for the Mirror implementation. These should be set with configuration and injected at runtime.

pferrel: don't bother with injection now, just asking.

PR #59, opened 2018-03-21
java sdk ssl fix

qqmbr4k:
Removed usage of HARNESS_SSL_CONFIG_PATH on both client and server. Made TLS works with just HARNESS_SSL_ENABLED=true.


pferrel:
The problem was to remove the use of akka-ssl.conf to use in "no-ssl" and "with-ssl" modes. The default mode is no-ssl, this should work with default `harness-env` settings. 

Are you saying that some form of "with-ssl" will work with only a change to `harness-env`? Maybe by using the provided `.pem` and `.jks`?

Can you describe what changes to config would enable "with-ssl"?

java-sdk/harness.pem
```diff
@@ -0,0 +1,23 @@
...
+0EuP3mzM4IsMb30LFXF5AgMBAAGjYjBgMBMGA1UdJQQMMAoGCCsGAQUFBwMBMA4G
+A1UdDwEB/wQEAwIFoDAaBgNVHREEEzARgglsb2NhbGhvc3SHBH8AAAEwHQYDVR0O
+BBYEFJkaLX3ExGGhtSlIyRJJ1tgrxt15MA0GCSqGSIb3DQEBCwUAA4IBAQAg8DC8
+iK+yyRNRRq4OaTIpz7StCuIHvAahrc8ZoGR1x5WPxYlXOFRY6RK3vZZAF541ruda
+ppVrj6sopO8BRgwjPpgcZfe4e9I4nYkPUystw/oDAdldxdIb81hgy3Ollmf4SiFh
+pWjUpjhNCZeXNfDt1nZ+QUvYFn3A9MyI3xeY3bRVCPqPvbmR1IoY+Ac+X5zm1mcE
+OVN5EDffyYAnui6JzTqVGZq1ts7PeB4UyVcqqEcV9ANhOzY2FD42OAZTwHa44hWm
+Q6VDZOPpCcCOoj4UsD2jbcsAW7ax8wJqjEcDvxiriiBd2VvV2aSOXfY3PEsPN0hb
+Sr71OrAYMmFVMlnP
+-----END CERTIFICATE-----
```
pferrel: What is this for?

pferrel:
If there is a way to test SSL, using an integration test, how would we do that?

pferrel:
I'll merge this since I have to update tests and will be adding some travis/ci tests but would like comments about any possible TLS tests.

andreyshikov:
@pferrel please do not merge right now, let me answer your questions...

java-sdk/harness.pem
andreyshikov: default certificate

andreyshikov:
> If there is a way to test SSL, using an integration test, how would we do that?

> Can you describe what changes to config would enable "with-ssl"?

Just `HARNESS_SSL_ENABLED=true` for default SSL setup. And also HARNESS_SERVER_CERT_PATH and maybe HARNESS_EXTERNAL_ADDRESS should be set properly for CLI.

pferrel:
looks like TLS + localhost does not work with Python SDK + CLI. It should but as long as the external address version of Python SDK + CLI + TLS works with a different cert this is ok for now. We will figure out localhost + TLS later since it is only useful for testing.

pferrel:
I assume TLS works from localhost now?

PR #60, opened 2018-03-27
Fixed python-sdk's TLS support

qqmbr4k:


pferrel:
Does this work with localhost so all you have to do is define TLS enabled and all the pointers to jks and pem work in TLS mode?

andreyshikov:
@pferrel HARNESS_SSL_ENABLED=true and HARNESS_SERVER_CERT_PATH=<java-sdk's harness.pem> and HARNESS_EXTERNAL_ADDRESS=localhost should be set for harness CLI. Java SDK works out of the box as usual.

PR #61, opened 2018-04-01
Feature/harness import

pferrel:
This is not working. 

`harness import /path/to/event/files/directory` is parsed correctly and the call to the SDK seems correct but the SDK fails. 

We are using a query with POST/update to do 2 things
 - `harness update /path/to/engine/json/config/file` which works fine
 - `harness import /path/to/event/files/directory` which dies in the POST either in the python SDK or is rejected by the server, not sure which. 

to test `import` use any contextual bandit engine and event file in json.


java-sdk/example/data/test_cb_mod.json
```diff
@@ -0,0 +1,19 @@
...
+  "modelContainer": "/Users/pat/big-data/harness/models/",
+  "algorithm":{
+    "maxIter": 100,
+    "regParam": 0.0,
+    "stepSize": 0.1,
+    "bitPrecision": 24,
+    "comment": "modelContainer is output an should be a full directory path on your local fs",
+    "maxClasses": 3
+  }
+}
```
pferrel, line 19: This is a test engine config that can be used with `harness update test_cb_mod.json` but requires there already be an engine of the same id: `test_cb`

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -122,6 +122,17 @@ class MongoAdministrator extends Administrator with JsonParser with Mongo {
...
+    engines.get(engineId).map { existingEngine =>
+      logger.trace(s"Importing a batch of events into engine: ${engineId} from $inputPath")
+      existingEngine.init(inputPath, deepInit = false).andThen(_ => Valid(
+        """{
+          |  "comment":"New events imported"
+          |}
+        """.stripMargin))
+    }.getOrElse(Invalid(WrongParams(s"Unable to import to Engine: $engineId}, the engine does not exist")))
+  }
+
```
pferrel, line 135: this code isn't finished yet because we never get here.

rest-server/bin/engines.py
```diff
@@ -25,7 +25,7 @@
     # print("Engine-id: " + engine_id)
     # print("Json config: \n" + str(config))
     try:
-        res = engine_client.update(engine_id=engine_id, data=config)
+        res = engine_client.update(engine_id=engine_id, input_path=args.input, update_config=True, data=config)
```
pferrel: the API to engine_client.update(...) has changed to support 2 variants, specified by the query part of the URL

rest-server/bin/engines.py
```diff
@@ -47,6 +47,18 @@
...
+    try:
+        #  def update(self, engine_id, input_path=None, update_config=False, data={}):
+        res = engine_client.update(engine_id=engine_id, input_path=args.input, update_config=False, data={})
+        print_success(res, 'Importing to engine: {}\n'.format(engine_id))
+    except HttpError as err:
+        print_failure(err, 'Error importing to engine-id: {} from {}\n'.format(engine_id, args.input))
+    #  else:
+    #      print_failure(None, "Error: no input for import command.")
+
 elif args.action == 'delete':
```
pferrel: this seems to work ok, but the exception is always throw:

Error importing to engine-id: test_cb from /Users/pat/big-data/harness/test_cb_backup
 Request entity expected but not supplied

I'm not sure what is throwing the exception originally, the server or the python SDK?

rest-server/bin/harness-engine-import
```diff
@@ -7,35 +7,32 @@
 
 while [ -n "$1" ]; do
     case "$1" in
-        *) RESOURCE_ID=$1
-            FILENAME=$2
-            #echo "Filename: ${FILENAME}"
+        *) resource_id=$1
+            input_path=$2
+            echo "Engine id: ${resource_id} filename: ${input_path}"
             ;;
```
pferrel: moving to the convention of small case bash vars when they are never kept in the env.

rest-server/core/src/main/scala/com/actionml/core/template/Algorithm.scala
```diff
@@ -30,17 +30,17 @@ import com.typesafe.scalalogging.LazyLogging
   */
 abstract class Algorithm[Q, R] extends JsonParser with LazyLogging {
   var engineId: String = _
-  var modelPath: String = _
+  var modelPath: String = _  // optional place in some filesystem to persist the model, a file path
 
   def init(json: String, engine: Engine): Validated[ValidateError, Boolean] = {
     engineId = engine.engineId
-    modelPath = engine.modelContainer // optional place in some filesystem to persist the model
+    modelPath = engine.modelContainer + engineId //todo: should the Engine even know about the modelContainer?
```
pferrel, line 37: fixed a bug in default naming of model files

rest-server/core/src/main/scala/com/actionml/core/template/Engine.scala
```diff
@@ -108,6 +108,11 @@ abstract class Engine extends LazyLogging with JsonParser {
     Valid( true )
   }
 
+  def batchInput(inputPath: String): Validated[ValidateError, Boolean] = {
+    mirroring.map(_.importEvents(this, inputPath))
+    Valid( true )
+  }
```
pferrel: this will do the work of import but we don't get here yet

rest-server/server/src/main/scala/com/actionml/router/service/EngineService.scala
```diff
@@ -55,6 +55,10 @@ class EngineServiceImpl(implicit inj: Injector) extends EngineService{
       log.info(s"Update existing engine, ${engineJson}")
       sender() ! admin.updateEngine(engineJson).map(_.asJson)
 
+    case UpdateEngineWithImport(engineId, inputPath) =>
+      log.info(s"Update existing engine by importing, ${inputPath}")
+      sender() ! admin.updateEngineWithImport(engineId, inputPath).map(_.asJson)
+
```
pferrel, line 61: we don't get this far

python-sdk/harness/__init__.py
```diff
@@ -318,18 +318,35 @@ def async_create(self, data):
...
+            print("Query for URL: {}".format(query))
+            # data = {"dummy": "data"}
+
+        path = self._add_segment(engine_id)
+        new_path = self._add_get_params(path, **query)
+        print("Fully constructed path: {}".format(new_path))
+        print("Data supplied: {}".format(data))
+
+        request = AsyncRequest("POST", new_path, **data)
+
```
pferrel: this is the root of the problem. The AsynchRequest does not seem to be created correctly

I was planning to do make the `import` REST API: 
`/engines/<engine-id>?import_path=<some URL Encoded path` 
with an empty body and the REST API for `update`:
 `/engines/<engine-id>?updage_config=true` with the new config in the body in json

pferrel:
@andreyshikov I can't seem to get the SDK to communicate with harness for `harness import` though `harness update` works, both using `POST /engines/<engine-id>...` The URL query params are slightly different and the body is empty in one case and hase JSON config in the other case.

Can you take a look at it, I can't see what I'm doing wrong.

java-sdk/example/data/test_cb_mod.json
pferrel: not important, you can ignore

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
pferrel: This is working now when all parameters are supplied

pferrel:
the integration-test is still not working correctly. It will complete if you run `./integration-test.sh -s` to skip restarts but the results are in error.

I suspect `harness stop` is causing a quiet crash or hang so that the next `harness start` is competing for the same port or starts up in some bad way because I often see 2 harness processes running at the same time. It might be VW but suspect that it is more likely to be the actor system used in the CBAlgorithm. @andreyshikov can you look at actor logic here:  https://github.com/actionml/harness/blob/feature/harness-import/rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scalahttps://github.com/actionml/harness/blob/feature/harness-import/rest-server/templates/src/main/scala/com/actionml/templates/cb/CBAlgorithm.scala

Do we need hook `kill` and do something to signal engines to shutdown then wait for them to finish?

rest-server/bin/harness-stop
```diff
@@ -3,15 +3,50 @@
...
 if [ -e ${PIDFILE} ]; then
     #cat ${PIDFILE} | xargs kill
     pid=`cat ${PIDFILE}`
     echo -e "${GREEN}Stopping Harness Server on pid: ${pid} ${NC}"
-    kill -9 ${pid} # Todo: this should not use -9, find why Harness is sometimes not stopping with only kill pid
+    #kill ${pid} # Todo: this should not use -9, find why Harness is sometimes not stopping with only kill pid
+    killwait ${pid}
+    #sleep ${sd_delay}
+
+    h=`jps | grep Main | wc -l`
```
andreyshikov, line 44: I usually use `grep actionml` for this. But Main is ok too

qqmbr4k (approved):

andreyshikov:
```while (!Files.deleteIfExists(Paths.get(modelPath))) {}```

can the `modelPath` file be opened by VW? In such case line 131 can hang

PR #62, opened 2018-04-08
Feature/async mongo 2.0

pferrel:


pferrel:
great, I'll do some testing. Once we get this passing tests and add some for the CLI this will be a good v0.2.0 candidate. 

We still need to look at `init`, right? Is it sync or async in this version?

andreyshikov:
> We still need to look at init, right? Is it sync or async in this version?

@pferrel Everything is synchronous now, I'll try to make 'not admin' part asynchronous where possible.

pferrel:
I think async is not a good idea and actually think there are some places where sync is not safe in a multi threaded situation.

Also I've found at least one logic error that I'll comment on so maybe we can talk about it.

rest-server/templates/src/main/scala/com/actionml/templates/navhinting/NavHintingDataset.scala
```diff
@@ -67,40 +65,37 @@ class NavHintingDataset(engineId: String) extends Dataset[NHEvent](engineId) wit
...
-              if (updatedJourney.nonEmpty) { // existing Journey so updAte in place
-                activeJourneysDAO.save(updatedJourney.get)
-                Valid(true)
-              } // else the first event for the journey is a conversion so ignore
-            } else { // no persisted journey so create it
-              activeJourneysDAO.insert(
-                Journey(
-                  event.entityId,
-                  Seq(JourneyStep(event.targetEntityId, DateTime.parse(event.eventTime)))
+          Await.result(activeJourneysDAO.find("_id" -> event.entityId).flatMap { unconvertedJourney =>
```
pferrel, line 73: There is a logic error here. In the develop branch `findOnById` return `None` when the user-id, which is the key for this doc, does not exist. This means I will create a new record but the way you wrote it causes an exception to be thrown. Notice the `if (unconvertedJourney.nonEmpty)` logic above. 

We could add a method to the DAO or do this some other way. BTW this seems unsafe because the queue I create might be created by another thread even if using sync DB calls, right? There are other cases where I read the DB, do some complex computation to create a new doc then upsert it. These are all unsafe, right?

Also the `.save` is the same as `upsert` semantically, I believe. And it is more clear so maybe we should use that word for the function.

PR #67, opened 2018-05-10
Fixed integration test:

qqmbr4k:
* removed joda time from the models
* added some exception handlers to the routes
* changed removing engines from actor ask (? method) to direct method call (to avoid timeout exceptions that hide original ones)
* added some shutdown hooks to the jvm and actor systems
* slightly changed NHNavEvent format to match json events
* added mongo client close on shutdown
* other minor fixes
* removed integration-test “fix” with restarts
* upgraded java-sdk to newer versions

### DIFF
--- a/java-sdk/example/data/test_cb.json
+++ b/java-sdk/example/data/test_cb.json
@@ -5,7 +5,7 @@
   "comment": "add sharedDBName for sharing a DBs for Users, should be named with lowercase and underscore separator, aka snake case",
   "sharedDBName": "harness_test_shared_users",
   "comment": "modelContainer is output an should be a full directory path on your local fs",
-  "modelContainer": "/Users/pat/big-data/harness/models/",
+  "modelContainer": "/tmp/harness/models/",
   "algorithm":{
     "maxIter": 100,
     "regParam": 0.0,
--- a/rest-server/bin/harness-start
+++ b/rest-server/bin/harness-start
@@ -15,14 +15,14 @@ if [[ "$1" = "-f" ]] || [[ ! -f ${PIDFILE} ]]; then
         echo -e "${CYAN}Trying to stop Harness from existing PID file before restarting${NC}"
         harness-stop
     fi
-    exec ${HARNESS_HOME}/bin/main \
+    exec ${HARNESS_HOME}/bin/main -v \
     -Dlogback.configurationFile=${HARNESS_LOG_CONFIG} \
     -DHARNESS_LOG_PATH=${HARNESS_LOG_PATH} \
     -Drest-server.host=${REST_SERVER_HOST} \
     -Drest-server.port=${REST_SERVER_PORT} \
     -Dmongo.host=${MONGO_HOST} \
     -Dmongo.port=${MONGO_PORT} \
-    <&- > /dev/null 2>&1 &
+    <&- > /tmp/harness.log 2>&1 &
     echo $! > ${PIDFILE}
     if [[ $? == 0 ]]; then
         echo -e "${CYAN}Started Harness with PID: $(cat ${PIDFILE})${NC}"
--- a/rest-server/engines/src/main/scala/com/actionml/engines/cb/CBAlgorithm.scala
+++ b/rest-server/engines/src/main/scala/com/actionml/engines/cb/CBAlgorithm.scala
@@ -310,7 +310,6 @@ object CBAlgorithm extends LazyLogging {
     } catch {
       case e: IOException => logger.error(s"Can't destroy $this", e)
       case e: TimeoutException => logger.error(s"Error unable to delete the VW model file for $engineId at $modelPath in the 3 second timeout.", e)
-      case e => logger.error(s"Can't destroy algorithm for $engineId (created with params $params) - unknown error", e)
     }
   }
 
--- a/rest-server/engines/src/main/scala/com/actionml/engines/cb/CBAlgorithm.scala
+++ b/rest-server/engines/src/main/scala/com/actionml/engines/cb/CBAlgorithm.scala
@@ -17,7 +17,7 @@
 
 package com.actionml.engines.cb
 
-import java.io.File
+import java.io.{File, IOException}
 import java.time.temporal.ChronoUnit
 import java.time.{Duration, Instant, LocalDateTime, OffsetDateTime}
 import java.util.concurrent.TimeoutException
@@ -107,7 +107,7 @@ class CBAlgorithm(resourceId: String, dataset: CBDataset)
       Valid(true)
     } catch {
       case e: NoSuchElementException =>
-        logger.error(s"Training triggered on non-existent group: $groupName Initialize the group before sending input.")
+        logger.error(s"Training triggered on non-existent group: $groupName Initialize the group before sending input.", e)
         Invalid(ValidRequestExecutionError(s"Input to non-existent group: $groupName Initialize the group before sending input."))
     }
   }
@@ -115,19 +115,6 @@ class CBAlgorithm(resourceId: String, dataset: CBDataset)
   override def predict(query: CBQuery): CBQueryResult = {
     // todo: isDefinedAt is not enough to know there have been events
     if(dataset.usageEventGroups isDefinedAt query.groupId) getVariant(query) else getDefaultVariant(query)
-  }
-
-  override def destroy(): Unit = {
-    try { Await.result(
-      actors.terminate().andThen { case _ =>
-        if (vw != null.asInstanceOf[VWMulticlassLearner]) vw.close()
-      }.map { _ =>
-        if (Files.exists(Paths.get(modelPath)) && !Files.isDirectory(Paths.get(modelPath)))
-          while (!Files.deleteIfExists(Paths.get(modelPath))) {}
-      }, 3 seconds) } catch {
-      case e: TimeoutException =>
-        logger.error(s"Error unable to delete the VW model file for $engineId at $modelPath in the 3 second timeout.")
-    }
   }
 
   def createVW(params: CBAlgoParams): VWMulticlassLearner = {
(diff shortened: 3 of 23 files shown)


java-sdk/example/data/test_cb.json
```diff
@@ -5,7 +5,7 @@
   "comment": "add sharedDBName for sharing a DBs for Users, should be named with lowercase and underscore separator, aka snake case",
   "sharedDBName": "harness_test_shared_users",
   "comment": "modelContainer is output an should be a full directory path on your local fs",
-  "modelContainer": "/Users/pat/big-data/harness/models/",
```
andreyshikov, line 8: @pat if you don't mind

rest-server/bin/harness-start
```diff
@@ -15,14 +15,14 @@ if [[ "$1" = "-f" ]] || [[ ! -f ${PIDFILE} ]]; then
...
-    exec ${HARNESS_HOME}/bin/main \
+    exec ${HARNESS_HOME}/bin/main -v \
     -Dlogback.configurationFile=${HARNESS_LOG_CONFIG} \
     -DHARNESS_LOG_PATH=${HARNESS_LOG_PATH} \
     -Drest-server.host=${REST_SERVER_HOST} \
     -Drest-server.port=${REST_SERVER_PORT} \
     -Dmongo.host=${MONGO_HOST} \
     -Dmongo.port=${MONGO_PORT} \
-    <&- > /dev/null 2>&1 &
+    <&- > /tmp/harness.log 2>&1 &
```
andreyshikov, line 25: I'd like to write it somewhere if this is not a problem for ops and disk space @pferrel

rest-server/engines/src/main/scala/com/actionml/engines/navhinting/NavHintingDataset.scala
```diff
@@ -84,7 +85,7 @@ class NavHintingDataset(engineId: String, store: Store)(implicit ec: ExecutionCo
               activeJourneysDAO.insert(
                 Journey(
                   event.entityId,
-                  Seq(JourneyStep(event.targetEntityId, OffsetDateTime.parse(event.eventTime)))
+                  Seq(JourneyStep(event.targetEntityId, OffsetDateTime.parse(event.eventTime))) // TODO: what to do with the parse error?
```
andreyshikov, line 88: @pferrel what do you think about fallback here?

rest-server/engines/src/main/scala/com/actionml/engines/navhinting/NavHintingDataset.scala
```diff
@@ -213,6 +214,7 @@ case class NHNavEventProperties(
 
 case class NHNavEvent(
     event: String,
+    entityType: String,
```
andreyshikov, line 217: added this because json events are different

rest-server/engines/src/main/scala/com/actionml/engines/cb/CBAlgorithm.scala
```diff
@@ -300,10 +287,33 @@ class CBAlgorithm(resourceId: String, dataset: CBDataset)
...
+          if (Files.exists(modelPath) && !Files.isDirectory(modelPath)) {
+            logger.info(s"Deleting file ${modelPath} for vw $vw")
+            while (!Files.deleteIfExists(modelPath)) {logger.info(s"trying to delete ${modelPath}")}
+          } else logger.info(s"$vw has no file to delete")
+        }, 2 minutes)
+      logger.info(s"Algorithm for $engineId (created with params $params) was destroyed")
+    } catch {
+      case e: IOException => logger.error(s"Can't destroy $this", e)
+      case e: TimeoutException => logger.error(s"Error unable to delete the VW model file for $engineId at $modelPath in the 3 second timeout.", e)
+      case e => logger.error(s"Can't destroy algorithm for $engineId (created with params $params) - unknown error", e)
```
pferrel: this is dangerous, it catches all errors, some should be fatal

andreyshikov: yeah, I'll remove it

java-sdk/example/data/test_cb.json
pferrel: if we add this then, since they are only used for int tests, the tests should check they exist or create them

pferrel: @andreyshikov

rest-server/engines/src/main/scala/com/actionml/engines/cb/CBAlgorithm.scala
pferrel: @andreyshikov

java-sdk/example/data/test_cb.json
andreyshikov: @pferrel I don't get it - maybe misplaced comment. I'm talking about replacement of `/Users/pat/big-data/harness/models/` with `/tmp/harness/models`

rest-server/server/src/main/scala/com/actionml/HarnessServer.scala
```diff
@@ -10,30 +10,45 @@ import com.actionml.router.service._
...
-object HarnessServer extends App with AkkaInjectable{
+object HarnessServer extends App with AkkaInjectable with LazyLogging {
 
   implicit val injector = new BaseModule
 
   inject[RestServer].run()
 
+  sys.addShutdownHook {
+    logger.info("Shutting down Harness Server")
+    MongoStorage.close
```
pferrel, line 35: @andreyshikov this makes a hard dependency on mongo, would be nice to avoid somehow

rest-server/server/src/main/scala/com/actionml/router/http/routes/EnginesRouter.scala
```diff
@@ -105,10 +108,13 @@ class EnginesRouter(implicit inj: Injector) extends BaseRouter with Authorizatio
...
 
+  private val admin = inject[Administrator]
   private def deleteEngine(engineId: String)(implicit log: LoggingAdapter): Route = {
     log.info("Delete engine: {}", engineId)
+    import io.circe.syntax._
+
     completeByValidated(StatusCodes.OK) {
-      (engineService ? DeleteEngine(engineId)).mapTo[Response]
+      Future.successful(admin.removeEngine(engineId).map(_.asJson))
     }
```
pferrel, line 118: @andreyshikov are we sure this is needed?

rest-server/server/src/main/scala/com/actionml/HarnessServer.scala
andreyshikov: yeah, maybe later we'll make a "registry" with closable callbacks but for now, it looks like a backlog task

rest-server/server/src/main/scala/com/actionml/router/http/routes/QueriesRouter.scala
```diff
@@ -47,4 +48,10 @@ class QueriesRouter(implicit inj: Injector) extends BaseRouter with Authorizatio
       (queryService ? GetPrediction(engineId, query.toString())).mapTo[Response]
     }
   }
+
+  private def exceptionHandler(log: LoggingAdapter) = ExceptionHandler {
+    case e =>
+      log.error("Internal error", e)
+      complete(StatusCodes.InternalServerError)
+  }
```
pferrel, line 56: what are the routing changes for?

rest-server/server/src/main/scala/com/actionml/router/http/routes/EnginesRouter.scala
andreyshikov: removing engines takes more then default ask timeout on my machine, so it make sense for me to get actual errors, not timeout exceptions.... But yes, I can remove it also

rest-server/server/src/main/scala/com/actionml/router/http/routes/QueriesRouter.scala
andreyshikov: exception handling. It wraps exceptions to the InternalServerError response codes

rest-server/engines/src/main/scala/com/actionml/engines/navhinting/NavHintingDataset.scala
```diff
@@ -213,6 +214,7 @@ case class NHNavEventProperties(
...
     entityId: String,
```
pferrel, line 218: @andreyshikov is this required?

andreyshikov: I suppose https://github.com/actionml/harness/blob/develop/java-sdk/example/data/nh-pferrel-events-urls.json

rest-server/server/src/main/resources/logback.xml
```diff
@@ -3,7 +3,7 @@
 
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%green(%d{HH:mm:ss.SSS}) %highlight(%-5level) %green(%-17logger{0}) - %msg%n</pattern>
+            <pattern>%d{HH:mm:ss.SSS} %-5level %-17logger{0} - %msg%n</pattern>
```
pferrel, line 6: Does this remove coloring from when viewing in terminal to tailing? I know the escape code junk up the logs but we always search or grep them anyway.

andreyshikov: yes, it removes "colored" logs

java-sdk/example/data/test_cb.json
pferrel: @andreyshikov Yes understood but those don't exist on any machine and since these are used only for tests, then the tests can create them if they don't exist

andreyshikov: I'm not sure it will be created, I created them myself on my laptop

pferrel: @andreyshikov notice that `.../harness/models` and `.../harness/mirrors` are used in every test engine.json so after any reboot or on each new install they will have to be created, why not do this in the tests so the test will just work out of the box, saving us countless hand tasks over time.

andreyshikov: @pferrel do you mean I should add `mkdir -p /tmp/harness/models /tmp/harness/mirrors` to the cb test.sh?

PR #68, opened 2018-05-26
Release/0.2.0 rc2

pferrel:
added a requested feature from XyXon to respond 404 to a $delete fo a model where the model does not exist.

This is not strictly best practice as I understand REST but we aren't treating a model as a first class resource like an Event so it's a gray area.


pferrel:
wrong diff, want to merge with develop

PR #69, opened 2018-05-26
Release/0.2.0 rc2

pferrel:
added a 404 response when $delete is for a non-existent model.

Not strictly correct REST but we aren't treating models like a first class Resources like Events


pferrel:
@andreyshikov can you check the changes to make sure I did this right? I need to deploy it Sunday night my time.

BTW is looks like release/0.2.0-RC1 is not up to date with what is on XyXon. Where is that code?

rest-server/engines/src/main/scala/com/actionml/engines/navhinting/NavHintingDataset.scala
```diff
@@ -106,9 +106,16 @@ class NavHintingDataset(engineId: String, store: Store)(implicit ec: ExecutionCo
               Valid(event)
             case "model" =>
               logger.trace(s"Dataset: ${engineId} removing model for conversion-id: ${event.entityId}")
-              navHintsModels.removeOneById(event.entityId) // todo: does this throw an exception if it fails to find?
-              store.removeCollection(event.entityId) // todo: does this throw an exception if it fails to find?
-              Valid(event)
+              try {
+                navHintsModels.removeOneById(event.entityId) // todo: does this throw an exception if it fails to find?
```
andreyshikov, line 110: yes, it throws a runtime exception if no entry is found. And all sync methods can throw the timeout exception

rest-server/engines/src/main/scala/com/actionml/engines/navhinting/NavHintingDataset.scala
```diff
@@ -106,9 +106,16 @@ class NavHintingDataset(engineId: String, store: Store)(implicit ec: ExecutionCo
...
+                store.removeCollection(event.entityId) // todo: does this throw an exception if it fails to find?
```
andreyshikov, line 111: it also throws a runtime exception if something went wrong

rest-server/engines/src/main/scala/com/actionml/engines/navhinting/NavHintingDataset.scala
```diff
@@ -106,9 +106,16 @@ class NavHintingDataset(engineId: String, store: Store)(implicit ec: ExecutionCo
...
+                Valid(event)
+
+              } catch {
+                case _ => //todo: @andrey java.lang.RuntimeException is thrown by sync dao, this seems to generic
+                  logger.error(s"Cannot $$delete non-existent model: ${event.entityId}")
+                  Invalid(ResourceNotFound(s"Cannot $$delete non-existent model: ${event.entityId}"))
```
andreyshikov, line 117: I suppose this behavior should be added to CBEngine too

qqmbr4k (dismissed):

PR #71, opened 2018-07-05
Feature/advanced dao filters

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/store/DaoQuery.scala
```diff
@@ -0,0 +1,30 @@
...
+case class DaoQuery(offset: Int = 0, limit: Int = Int.MaxValue, orderBy: Option[OrderBy] = None, filter: Seq[(String, Any)] = Seq.empty)
+
+object Ordering extends Enumeration {
+  type Ordering = Value
+  val asc = Value("asc")
+  val desc = Value("desc")
+}
+
+case class OrderBy(ordering: Ordering.Ordering, fieldNames: String*)
+
```
pferrel, line 30: can I ask for 100 events with name="buy" AND userId = "some-user" ordered by the event.date, descending? Can you show an example assuming event is:

```
case class Event(_id: ObjectId, userId: String, name: String, date: Datetime)
```

Use whatever datetime type you want. 

Also doesn't `_id`, `name`, `userId`, and `date` need to be indexed to make this query fast? How do we specify what is indexed? We can assume `_id` is because it is the default object id.

pferrel: Never mind, I can see how to encode the query but still have a question about indexes.

pferrel:
ok, looks good but still have a question about indexes for ordering and filterable attributes

PR #72, opened 2018-07-11
Added spark support

pferrel:


rest-server/core/src/main/scala/com/actionml/core/spark/SparkConfigSupport.scala
```diff
@@ -0,0 +1,47 @@
...
+  def createSparkContext(config: String): Validated[ValidateError, SparkContext] = {
+    parseAndValidate[SparkConfig](config).andThen { sparkConfig =>
+      try {
+        val conf = new SparkConf()
+          .setMaster(sparkConfig.master)
+          .setAppName(sparkConfig.appName)
+        conf.set("spark.mongodb.input.uri", "mongodb://localhost/") // todo take it from file config or from incoming json config?
+        conf.set("spark.mongodb.input.database", sparkConfig.database)
+        conf.set("spark.mongodb.input.collection", sparkConfig.collection)
+        Valid(new SparkContext(conf))
```
pferrel: Many of these will be optional, the defaults supplied by Spark itself, not our own defaults. This means that if the config does not exist in the JSON, it should not be set at all.

Also there will be key/value pairs that should be set that are not understood by Spark because they are params to being passed to Spark libs like Elaticsearch, or Mongo's Spark lib. These should be of the form

```
conf.set(someKeyString, someValueString) // at least I know only of strings passed in
```

For example:

```
conf.set("spark.es.nodes", "node1,node2")
```

There will be no way for `SparkConfigSupport` to parse these arbitrary key/value pairs. They must be taken from the JSON but cannot be part of a `SparkConfig ` case class in `SparkConfigSupport`

rest-server/core/src/main/scala/com/actionml/core/spark/SparkLambdaAlgorithm.scala
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+
+
+  private def processPartition: Iterator[Document] => String = i => {
+    i.foldLeft("")(_ + _)
+  }
+
+  private def checkResult[T](a: Array[T]): Boolean = a.nonEmpty && a.foldLeft("")(_ + _).trim.nonEmpty
+}
```
pferrel: This should make the context an implicit val so it can be used in any Algorithm code. So take each of your methods and move them to `SparkScaffoldAlgorithm` and see if they works there. In other words ideally this is a trait that any Algorithm class can mixin, just like the `LambdaAlgorithm` trait but with extra support for the `sc`, which is typically what the sparkContext is called. 

The way we use this should be the same as LambdaAlgorithm, as a mixin usable by any Algorithm to add support for creating a context. IMO the Algo cannot be a concrete class that gets newed. Otherwise config parsing and context creation would have to be duplicated in every spark specific Algorithm.

rest-server/core/src/main/scala/com/actionml/core/spark/SparkScaffoldEngine.scala
```diff
@@ -0,0 +1,46 @@
...
+
+package com.actionml.core.spark
+
+import cats.data.Validated
+import com.actionml.core.engine.Engine
+import com.actionml.core.validate.ValidateError
+
+class SparkScaffoldEngine extends Engine {
+  private val lambdaAlgo = new SparkLambdaAlgorithm
+
```
pferrel: I don't see it this way, it should ideally be 

```
private val ldaAlgo = new LDAAlgorithm with SparkLambdaAlgorithm
```

this should make an `implicit val sc: SparkContext = ...` 

Also the parameters are in the engine's JSON and so already available as a string, so we may not need to pass them in to `Algo.train`.

pferrel:
@andreyshikov not sure what you are doing here. Spark support should be independent of any Algo or Engine class, as we talked about. It should ideally be more like the Mongo client, in the sense that it is separate.

pferrel:
Oh, I see looking at the single commit. You have a SparkLambdaAlgo as a helper to get Spark Support. That is nice.

rest-server/data/ur.json
```diff
@@ -3,21 +3,22 @@
   "engineFactory": "com.actionml.engines.ur.UREngine",
   "sparkConf": {
     "master": "yarn",
+    "appName": "harness-spark-test-app",
```
pferrel: this should be derived from the engine-id so it is guaranteed to be unique across all jobs by all engines.

rest-server/data/ur.json
```diff
@@ -3,21 +3,22 @@
...
+    "mongo.database": "harness_meta_store",
```
pferrel: This is not needed is it, in any case it should not be defined here but discovered from harness, maybe by a call to the administrator? Why is it needed? Ideally it is private to the administrator.

rest-server/data/ur.json
```diff
@@ -3,21 +3,22 @@
...
+    "mongo.collection": "engines",
```
pferrel: same comments as `mongo.database` above

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
```diff
@@ -0,0 +1,75 @@
...
+ * limitations under the License.
+ */
+
+package com.actionml.core.spark
+
+import cats.data.Validated
+import cats.data.Validated.{Invalid, Valid}
+import com.actionml.core.validate._
+import com.mongodb.spark.MongoSpark
+import com.mongodb.spark.rdd.MongoRDD
```
pferrel: The Spark context creation should not be coupled to Mongo.

pferrel: Mongo and Elasticsearch clients should consume the context and even provide key/value pairs for creating the context but no method of the SparkContextSupport should assume Mongo or Elasticsearch.

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
```diff
@@ -0,0 +1,75 @@
...
+import org.json4s.jackson.JsonMethods._
+
+
+trait SparkContextSupport extends LazyLogging {
+  import com.actionml.core.spark.SparkContextSupport._
+
+  def createRdd(sc: SparkContext): MongoRDD[Document] = {
+    MongoSpark.load[Document](sc)
+  }
+
```
pferrel: this method looks like is should be in the mongo client, maybe an extension of the base client that supports Spark. leaving the base Mongo client free of the need for Spark dependencies.

There will be a similar requirement for Elasticsearch, it must write an RDD (or Dataframe) to ES but should not pollute the SparkContextSupport with ES dependencies. Again, maybe classes for `ESClient` and `SparkESClient extends ESClient` would be the way to go.

rest-server/core/src/main/scala/com/actionml/core/spark/SparkLambdaAlgorithm.scala
```diff
@@ -0,0 +1,48 @@
...
+import com.actionml.core.engine.LambdaAlgorithm
+import com.actionml.core.validate.{JsonParser, ValidRequestExecutionError, ValidateError}
+import org.apache.spark.SparkContext
+import org.bson.Document
+
+class SparkLambdaAlgorithm(sparkContextConfig: String) extends LambdaAlgorithm[String] with SparkContextSupport with JsonParser {
+
+  private lazy val validatedSparkContext: Validated[ValidateError, SparkContext] = createSparkContext(sparkContextConfig)
+
+  override def train(): Validated[ValidateError, String] = validatedSparkContext.andThen { sparkContext =>
```
pferrel: I think we talked about the fact that we should not need a special SparkAlgorithm, it should be possible to mix-in a trait to get Spark support, something like:

```scala
class URAlgorithm(...) extends LambdaAlgorith(...) with JsonParser() with SparkSupport(...)
...
```

This way there will be an implicit `sc` for any method that needs it. The params for the context should be created in the factory so they can be passed in to the constructors. 

BTW this is the basis of my bloody merge--doing all the init stuff in the factory.

rest-server/data/ur.json
andreyshikov: @pferrel I've changed appName to `engineId`
Collection (and database it belongs to) is needed by spark job to create RDD from it

pferrel: ok. I have struggled with what to name things like DBs, etc. We only have the engineId that is guaranteed to be unique but we can append things to this to make it more descriptive if needed. Like engineId+"training" or smth.

rest-server/data/ur.json
pferrel: Why is the `harness_meta_store` name needed? This should never be used outside the administrator.

pferrel:
ok this looks pretty good. Make sure to no assume the DB name is the same as the engineId. You should get the DB name from the Store. There are cases when the DB name is not the engine-id, where data is shared between Engines.

Also it seems like a reference to the context will be stored in the consuming class, this must be a var since the context will be modified often while the Engine is running.

PR #74, opened 2018-08-06
Feature/spark support

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
```diff
@@ -0,0 +1,65 @@
...
+  }.andThen { sparkConfig =>
+    try {
+      val dbUri = MongoStorage.uri
+      val conf = new SparkConf()
+        .setMaster(sparkConfig.master)
+        .setAppName(sparkConfig.appName)
+      conf.set("deploy-mode", "cluster")
+      conf.set("spark.mongodb.input.uri", dbUri)
+      conf.set("spark.mongodb.input.database", sparkConfig.database)
+      conf.set("spark.mongodb.input.collection", sparkConfig.collection)
```
pferrel, line 47: No, db name and collection name should come from Store and DAO, we do not want to proliferate unneeded config changes. "Convention over Configuration" will make config simpler.

rest-server/core/src/main/scala/com/actionml/core/spark/SparkMongoSupport.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import com.mongodb.spark.MongoSpark
+import com.mongodb.spark.rdd.MongoRDD
+import org.apache.spark.SparkContext
+import org.bson.Document
+
+
+trait SparkMongoSupport {
+
+  def createRdd(sc: SparkContext): MongoRDD[Document] = {
```
pferrel, line 28: if this returns a `MongoRDD` how is this generic? Anything that uses this is coupled to Mongo. Is this really necessary? What is special about an "MongoRDD"?

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoStore.scala
```diff
@@ -81,6 +81,8 @@ class MongoStorage(db: MongoDatabase, codecs: List[CodecProvider]) extends Store
 
   private val timeout = 5 seconds
   private def sync[A](f: => Future[A]): A = Await.result(f, timeout)
+
+  override def dbName: String = db.name
```
pferrel, line 85: This is ok if you never assume the dbname is the engine-id. There are cases when they are not the same but the dbname is always passed to the Store.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -71,6 +89,18 @@ class URAlgorithm(dataset: URDataset)
...
+  case class URAlgorithmParams(comment: String,
+                               esMaster: String,
+                               indexName: String,
+                               typeName: String,
+                               availableDateName: String,
+                               expireDateName: String,
+                               dateName: String,
+                               num: String)
+}
+
```
pferrel, line 103: wrong style, see https://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments for methods and https://docs.scala-lang.org/style/declarations.html for class declarations

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -32,11 +33,12 @@ import scala.language.reflectiveCalls
   *
   * @param engineId The Engine ID
   */
-class URDataset(engineId: String) extends Dataset[GenericEvent] with JsonParser {
-
+class URDataset(val engineId: String) extends Dataset[GenericEvent] with JsonParser {
+  override lazy val dbName: String = engineId
```
pferrel, line 37: no, error. The dbname is not always the engineId, you MUST get it from the Store or make the DAO know what store it uses and get it from the DAO.

This is an error since sometimes the dbname is NOT the enginId.

pferrel:
Does this do anything yet?

We need to import some events, store them in Mongo, then read them into a Spark RDD and process them before this is done enough to proceed.

pferrel:
didn't mean to close--oops

PR #75, opened 2018-08-06
Feature/elastic support

qqmbr4k:
it also contains spark-support branch


rest-server/core/src/main/scala/com/actionml/core/search/SearchSupport.scala
```diff
@@ -0,0 +1,38 @@
...
+  def createSearchClient(hosts: String*): SearchClient[T]
+}
+
+trait SearchClient[T] {
+  def close: Unit
+  def createIndex(indexName: String,
+                  indexType: String,
+                  fieldNames: List[String],
+                  typeMappings: Map[String, (String, Boolean)] = Map.empty,
+                  refresh: Boolean = false): Boolean
```
pferrel, line 34: This is a scala style violation

pferrel:
The ElasticSearchClient should be a singleton for the Engine if not all of Harness. Is this how it works? I didn't see where the client is used.

rest-server/core/src/main/scala/com/actionml/core/search/SearchSupport.scala
```diff
@@ -0,0 +1,38 @@
...
+  def deleteIndex(indexName: String, refresh: Boolean = false): Boolean
+  def refreshIndex(indexName: String): Unit
+  def search(query: String, indexName: String): Option[T]
+}
```
pferrel, line 38: The query should not be a string. It should probably be something a query builder creates with "must" "should", and "shouldNot" clauses among other things. The string should be constructed from these lists and produce the query string in the doc describing the ES changes from the old code in PIO.

pferrel: Or we leave this as something to fall back on if the builder is not robust enough.

PR #76, opened 2018-08-21
Moved spark context to the singleton object

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -69,42 +67,25 @@ class URAlgorithm[T] private (initParams: String, dataset: Dataset[T]) extends A
...
-
-
+    val defaults = Map(
+      "appName" -> dataset.engineId,
+      "deploy-mode" -> "cluster",
+      "spark.mongodb.input.uri" -> MongoStorage.uri,
+      "spark.mongodb.input.database" -> dataset.dbName,
+      "spark.mongodb.input.collection" -> dataset.collection
+    )
+    val result = Await.result(execute(myTrainFunction, initParams, defaults), Duration("5 minutes"))
```
pferrel: How would you use the execute to 
1) return the Valid or Invalid synchronous with the call to `train`
2) run process in the Future thread. 

It looks like `myTrainFunction` must take some iterator. This is not ideal, it would be better if a block of code `{some code}` could be specified that has access to all variables in the scope of the block. There should be no restrictions on input (all variables in the same scope as the block) or output (Unit is fine, output is generally in some service like a db or HDFS).

Packaging the input to `execute` as a named function with specific input/output does not seem ideal.

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
```diff
@@ -17,49 +17,74 @@
...
-}
 
-object SparkContextSupport {
-  private case class SparkConfig(master: String, appName: String, database: String, collection: String, properties: Map[String, String])
+  private case class SparkContextParams(config: String, defaults: Map[String, String])
+
+  private sealed trait SparkContextState
+  private case class Running(currentParams: SparkContextParams,
+                             currentPromise: Promise[SparkContext],
+                             otherPromises: Map[SparkContextParams, Promise[SparkContext]]) extends SparkContextState
```
pferrel, line 79: This does not follow the Scala Style Guide for indentation.

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
```diff
@@ -17,49 +17,74 @@
...
+import scala.util.Try
+
 
-trait SparkContextSupport extends LazyLogging {
-  import com.actionml.core.spark.SparkContextSupport._
+trait SparkContextSupport[A] {
+  this: SparkStorageSupport[A] with LazyLogging =>
+  import SparkContextSupport._
+
+  def execute[B](fn: Iterator[A] => B, config: String, defaults: Map[String, String])(implicit ca: ClassTag[A], cb: ClassTag[B]): Future[Array[B]] = {
```
pferrel: Ideally this should execute a block of code in the scope of the calling function. To the Scala compiler this would be an anonymous function whose input is whatever variables in scope that it references.  Can we do something like:

```
sparkContext.execute { 
  ...
}
```

where the block is executed in a new thread but has access to all variables in the scope of the block and also has an implicit `sc`

rest-server/core/src/main/scala/com/actionml/core/spark/SparkMongoSupport.scala
```diff
@@ -18,15 +18,17 @@
...
-import com.mongodb.spark.rdd.MongoRDD
 import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
 import org.bson.Document
 
 
-trait SparkMongoSupport {
+trait SparkStorageSupport[T] {
+  def createRdd(sc: SparkContext): RDD[T]
+}
```
pferrel: ok but ideally wouldn't you pass in a DAO with a filter and ordering? Then the Abstract `SparkStorageSupport` would have enough abstract DAO params to execute the implementation of `createRdd`, right?

BTW some naming conventions: 1) I renamed Storage to Store 2) shouldn't this be `readRdd` since we'll eventually have `writeRdd` at least in the Elasticsearch case and probably for any Store.

andreyshikov: > ok but ideally wouldn't you pass in a DAO with a filter and ordering? Then the Abstract SparkStorageSupport would have enough abstract DAO params to execute the implementation of createRdd, right?

I see no support for filtering and ordering inside mongo-spark lib. So, I suppose, we shouldn't add those params to our API since our only backend doesn't support it.

Fixed naming.

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
andreyshikov: It's default idea formatting, I see scala style indentation on methods, not case classes. In case you want to use method conventions for case classes, we can use scalafmt plugin or something...

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
andreyshikov: we need an id for context. For now, it's config + defaults. I can change execute signature to
``` def execute[B](config: String, defaults: Map[String, String])(fn: Iterator[A] => B)(implicit ca: ClassTag[A], cb: ClassTag[B]): Future[Array[B]]``` and call will became 


```
execute(initParams, defaults) { case i: Iterator[Document] =>
        ???
    }``` is this ok?

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
andreyshikov: we should talk about execute API...

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
pferrel: you can change the values in IDEA

pferrel: ok, I put the setting jar in Slack

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
pferrel: resolved

rest-server/core/src/main/scala/com/actionml/core/spark/SparkMongoSupport.scala
pferrel: hmm, so the only thing it does is read the whole collection into an RDD? Then we filter in Spark. It is common enough that we can add that to our abstract API. You may want to curry the `sc` if you do an `rdd.filter`

pferrel:
if this branch is just Spark not Mongo then we can delete it now I think.

PR #77, opened 2018-08-23
Feature/elastic support

pferrel:
Comparing to develop only so we can comment on the EsClient


rest-server/core/src/main/scala/com/actionml/core/search/SearchSupport.scala
```diff
@@ -15,12 +15,24 @@
...
+trait SearchClient[T] {
+  def close: Unit
+  def createIndex(indexName: String,
+                  indexType: String,
+                  fieldNames: List[String],
+                  typeMappings: Map[String, (String, Boolean)] = Map.empty,
+                  refresh: Boolean = false): Boolean
+  def deleteIndex(indexName: String, refresh: Boolean = false): Boolean
+  def refreshIndex(indexName: String): Unit
+  def search(query: String, indexName: String): Option[T]
```
pferrel, line 37: need a way to maintain an Alias and an index pointed to by the alias. The UR uses aliases , when an index is created unique string is appended to the actual index name but an alias of the indexName -> indexName + some unique string (epoch time?). This way we can create a new index and get it indexed then hot swap the new index so the alias points to it, then delete the old index. This uses one atomic action to change the model being used for queries with no downtime.

pferrel: This swapping might be done automatically in some way, hiding the alias management. Also it is important to check for old dangling indexes that no alias points to or there may be cases where they aren't cleaned out and fill up Elasticsearch. Again using the engineId as the base Alias and as the building block for the real indexName is important.

pferrel: So in our search client an indexName is the alias name, what it points to is the actual index created. For the UR this will be when we write an RDD to Elasticsearch, it will have the real index name and  when it is done and the indexing is complete, the Alias will be swapped to point to the new index. From then on any query will go to the Alias and therefore to the new index.

pferrel:
Some comment on the EsClient

I suggest you structure these like DAO/MongoDao in a similar directory structure since it is likely we may have to support Solr, which has most of the same features as ES but different control commands.

Don't worry about Solr now, but we will be ready.

pferrel:
Also, I'm sure you know we need to support Spark writing of an ES index, perhaps as a separate trait because most use of ES should not require Spark or any imports from Spark.

andreyshikov:
looks outdated

PR #78, opened 2018-08-28
Refactored mongo-spark to support generics

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/spark/SparkMongoSupport.scala
```diff
@@ -17,18 +17,43 @@
...
-trait SparkStoreSupport[T] {
-  def readRdd(sc: SparkContext): RDD[T]
+
+trait SparkStoreSupport {
+  def readRdd[T](sc: SparkContext, codecs: List[CodecProvider])(implicit ct: ClassTag[T]): RDD[T]
+}
+
+trait SparkMongoSupport extends SparkStoreSupport {
+
+  override def readRdd[T](sc: SparkContext, codecs: List[CodecProvider] = List.empty)(implicit ct: ClassTag[T]): RDD[T] = {
```
pferrel: This should be in the DAO, shouldn't it? It should essentially be a `readMany` that returns an RDD, right? Allowing the consumer to pass query params like filters. This will be quite typical in real use.

pferrel: no we don't have time to refactr this into DAO

PR #79, opened 2018-09-03
Feature/elastic ur

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/search/SearchSupport.scala
```diff
@@ -17,22 +17,30 @@
...
   def createIndex(indexName: String,
                   indexType: String,
                   fieldNames: List[String],
                   typeMappings: Map[String, (String, Boolean)] = Map.empty,
                   refresh: Boolean = false): Boolean
   def deleteIndex(indexName: String, refresh: Boolean = false): Boolean
   def refreshIndex(indexName: String): Unit
-  def search(query: String, indexName: String): Option[T]
+  def search(query: SearchQuery, indexName: String): Seq[T]
 }
```
pferrel: need alias management, aliases are like symlinks we always query against an alias named after the engineId, but the query is executed on an index of the engineId+epoch-as-string. So later when we create as new index we first create this uniquely named index then swap the alias to point to it, then delete the old index. There is also login in there to look for dangling unreferenced indexes to clean out. 

look at the `hotSwap` method in the UR's EsClient. Given out batch updates to the indexes this aliasing functionality should probably be in these classes. I think that means there should be no knowledge of the actual index names outside these classes and we only pass in the alias, which will be the engineId.

pferrel: also I think there must be some type mappings. We don't need to mess with dates for now

https://github.com/actionml/universal-recommender/blob/master/src/main/scala/URAlgorithm.scala#L955

pferrel: This function writes to ES and swaps the alias to the new index, then deletes the old index. This make all incoming queries work continuously with no downtime. This, it now a very good factoring of the functionality though.

https://github.com/actionml/universal-recommender/blob/master/src/main/scala/EsClient.scala#L256

rest-server/core/src/main/scala/com/actionml/core/search/SearchSupport.scala
```diff
@@ -17,22 +17,30 @@
 
 package com.actionml.core.search
 
-import org.json4s.JValue
-
-import scala.concurrent.Future
-
 trait SearchSupport[T] {
-  def createSearchClient(hosts: String*): SearchClient[T]
```
pferrel, line 25: where does it get hosts now?

andreyshikov: config: elasticsearch.uri

rest-server/server/src/main/resources/reference.conf
```diff
@@ -22,3 +22,11 @@ auth {
   client-id = "harness"
   client-secret = "mwjmo4fwC9qFW081gm5vwsmAC4Z2747XEU8eeV6dzUzjW"
 }
+
+elasticsearch {
+  uri = "http://localhost:9200"
```
pferrel, line 27: URI should be in harness config, or env var, or provided here as a fallback. Like Mongo URI. This is for operation without Spark like queries etc. Anything needed for Spark should be generated from internal data or from engine.json `sparkConf`

The ES Rest host list is in the PIO+UR engine.json in the `sparkConf` since it is passed to Spark for `saveToEs`

rest-server/server/src/main/resources/reference.conf
```diff
@@ -22,3 +22,11 @@ auth {
...
+  auth {
+    username = ""
+    password = ""
+  }
```
pferrel: Auth should be disabled or using the same conf in harness-env.sh, the env, of fallback to the def here. Like Mongo.

With the exception of anything Spark needs, which should be derived from things known inside Harness already if possible. So if the SparkContext needs auth info it should be found from the `harness-env.sh` settings or in the env, ...

PR #81, opened 2018-09-12
DO NOT MERGE - just an experiment

qqmbr4k:


pferrel:
@andreyshikov 

I still don't understand why readRdd is not a function of the DAO? Both reference a type of database engine (Mongo in our current usage) a db name (known by the Store), and a collection name (known by the DAO) So it seem like we should be able to do:

```
  // the store now would have some kind of Spark Support mixed in and would be injectable
  // therefor it would return a Spark aware DAO implementation that would have the readRdd API
  val activeJourneysDao = store.createDao[URNavHintingEvent]("active_journeys")

  val journeysRdd = activeJourneysDao.readRdd()(sc)
```

This is a more simple and clear API and moves all implementation into the DAO where it belongs. There is no need for any code to know the db name and collection name now outside of the DAO.

Furthermore this allows us to use filters or more complex queries and implement them in the DAO but using Spark to do the filtering. Much nicer.

PR #84, opened 2018-09-19
Added request to check ES index existance

qqmbr4k:
this pull request is temporary against release/0.3.0-RC1

### DIFF
--- a/rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
+++ b/rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
@@ -30,7 +30,7 @@ import org.apache.http.impl.nio.client.HttpAsyncClientBuilder
 import org.apache.http.nio.entity.NStringEntity
 import org.apache.http.util.EntityUtils
 import org.apache.spark.rdd.RDD
-import org.elasticsearch.client.RestClient
+import org.elasticsearch.client.{Request, RestClient}
 import org.elasticsearch.client.RestClientBuilder.HttpClientConfigCallback
 import org.json4s.DefaultReaders._
 import org.json4s.jackson.JsonMethods._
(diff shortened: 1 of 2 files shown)


rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
```diff
@@ -212,6 +212,11 @@ class ElasticSearchClient[T] private (alias: String) extends SearchClient[T] wit
     oldIndexSet.foreach(deleteIndexByName(_, refresh = false))
   }
 
+  override def isIndexExists: Boolean = {
+    val request = new Request("HEAD", s"/$alias")
```
pferrel, line 216: what does this request do?

We may have

 - an alias pointing to an index 
 - and one or more dangling indexes that have no alias pointing to them
 - we may have an alias with no index it points to. 

All are bad state that we need to clean up.

when URNavHintingEngine.destroy is called rather than look for an index is might be better to handle the cleanup logic in the client so instead of doing:

```
if(es.isIndexExists) es.deleteIndex()
```

It might be better to have `es.deleteIndex()` do the check internally and do all cleanup needed so we really don't need to check if the index exists outside the client code. `deleteIndex` would only be called when all persistent data for the Engine including its metadata known by the administrator are cleaned out completely. We already clean the URNavDataset and the Administrator deletes metadata, this is all that is remaining.

pferrel:
See the comment @andreyshikov I thought there might be a better simpler API.

rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
andreyshikov: >what does this request do?

original task:
> I need an API to find if there is an index created

and then the description of a case with the train method

nothing about delete or cleanup. So looks like this PR implements the original task.

```deleteIndex``` removes all indexes with the alias ElasticSupport:98

PR #85, opened 2018-10-01
Feature/spark standalone

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingAlgorithm.scala
```diff
@@ -268,10 +268,10 @@ class URNavHintingAlgorithm private (engine: URNavHintingEngine, initParams: Str
       "spark.mongodb.input.collection" -> dataset.getIndicatorEventsCollectionName
     )
 
-    SparkContextSupport.getSparkContext(initParams, defaults).map { implicit sc =>
+    SparkContextSupport.getSparkContext(initParams, defaults, Array(classOf[URNavHintingEvent])).map { implicit sc =>
```
pferrel, line 271: What uses the Array of classes? Kryo? If so what classes need to be passed? This needs to be documented somewhere.

pferrel: for instance what if we had to have 2 readRdds from 2 DAO collections? This is the case for the full UR that we do not need for URNavHinting.

What other times do we need to include classes in this array? It is called `kryoClasses`. Ok but under what circumstances do we need to pass them in rather than simply inherit `Serializable`?

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingPreparator.scala
```diff
@@ -100,7 +100,7 @@ object URNavHintingPreparator extends LazyLogging with SparkMongoSupport {
         logger.info("EventName: " + eventName)
         // logger.info(s"first eventName is ${trainingData.actions.head._1.toString}")
         val ids = if (eventName == trainingData.actions.head._1.toString && trainingData.minEventsPerUser.nonEmpty) {
-          val dIDS = IndexedDatasetSpark(eventRDD, trainingData.minEventsPerUser.get)(sc)
+          val dIDS = IndexedDatasetSparkHelper(eventRDD, trainingData.minEventsPerUser.get)(sc)
```
pferrel, line 103: Huh? I was adding a constructor to IndexedDatasetSpark, a Mahout defined type.  Why change the name? It is not returning a "helper" it is returning an `IndexedDatasetSpark`

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingPreparator.scala
```diff
@@ -137,7 +137,7 @@ object URNavHintingPreparator extends LazyLogging with SparkMongoSupport {
 }
 
 /** Companion Object to construct an IndexedDatasetSpark from String Pair RDDs */
-object IndexedDatasetSpark {
+object IndexedDatasetSparkHelper {
```
pferrel, line 140: Same comment as above.

pferrel:
@andreyshikov after you look at my comments please merge this into develop. We need to keep develop more up-to-date. If someone asks for a special release like RC1 we can handle it with a release branch that is semi-long-lived as we did with RC1. But Develop should be our major merge-to branch, right?

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingPreparator.scala
andreyshikov: There is an import of IndexedDatasetSpark from mahout. And at this line, you hide it with this declaration. This situation confuses me and the compiler (there is a warning). So I think it's better to have separate utility object for creating IndexedDatasetSpark instances.

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingAlgorithm.scala
andreyshikov: kryo types should be set explicitly as a spark context setting. In case of URNavHintingAlgorithm there is a computation on cluster that requires serialization/deserialization of URNavHintingEvent. That's why URNavHintingEvent must be registered on spark context that way.

PR #87, opened 2018-10-03
readRDD (from mongo) added to DAO

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/spark/GenericMongoConnector.scala
```diff
@@ -0,0 +1,42 @@
...
+import com.actionml.core.store.backends.MongoStorage
+import com.mongodb.MongoClient
+import com.mongodb.client.MongoDatabase
+import com.mongodb.spark.{MongoClientFactory, MongoConnector}
+import com.typesafe.scalalogging.LazyLogging
+import org.bson.codecs.configuration.CodecProvider
+
+import scala.reflect.ClassTag
+
+
```
andreyshikov, line 29: this code is extracted to a separate file from SparkContextSupport.scala

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
```diff
@@ -68,16 +68,18 @@ object SparkContextSupport extends LazyLogging {
   /*
```
andreyshikov, line 68: this file doesn't contain anything new, just refactoring

rest-server/core/src/main/scala/com/actionml/core/spark/SparkStoreSupport.scala
```diff
@@ -0,0 +1,34 @@
+/*
```
andreyshikov: extracted interface to a separate file

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingAlgorithm.scala
```diff
@@ -64,8 +64,9 @@ class URNavHintingAlgorithm private (engine: URNavHintingEngine, initParams: Str
   with JsonParser {
 
   import URNavHintingAlgorithm._
+  import com.actionml.core.spark.SparkMongoSupport.syntax._
 
-  private var sparkContext: Validated[ValidateError, SparkContext] = _
+  private lazy val eventsDao = dataset.store.createDao[URNavHintingEvent](dataset.getIndicatorEventsCollectionName)
```
andreyshikov: @pferrel looks like that you were talking about ^^^

btw dataset.store doesn't look like a good idea, but store is usually a dataset member, so I'd like to discuss it with you

rest-server/core/src/main/scala/com/actionml/core/spark/SparkContextSupport.scala
```diff
@@ -68,16 +68,18 @@ object SparkContextSupport extends LazyLogging {
...
   Creates one context per jvm and one parameters set and gives promises for future contexts.
    */
-  def getSparkContext(config: String, defaults: Map[String, String], kryoClasses: Array[Class[_]] = Array.empty): Future[SparkContext] = {
-    val params = SparkContextParams(config, defaults: Map[String, String], kryoClasses)
+  def getSparkContext(config: String, appName: String, kryoClasses: Array[Class[_]] = Array.empty): Future[SparkContext] = {
+    val params = SparkContextParams(config, appName, kryoClasses)
     state.get match {
```
pferrel, line 73: Removal of defaults means we must do all changes to the config before getting the `SparkContext`? This seems good.

rest-server/core/src/main/scala/com/actionml/core/spark/SparkMongoSupport.scala
```diff
@@ -70,16 +58,20 @@ trait SparkMongoSupport extends SparkStoreSupport {
...
-  override def create(): MongoClient = new GenericMongoClient[T](host, codecs, ct)
-}
-
-class GenericMongoClient[T](host: String, codecs: List[CodecProvider], ct: ClassTag[T]) extends MongoClient(host) with LazyLogging {
-
-  override def getDatabase(databaseName: String): MongoDatabase =
-    super.getDatabase(databaseName).withCodecRegistry(MongoStorage.codecRegistry(codecs)(ct))
+object SparkMongoSupport {
+  object syntax {
+    implicit class DaoSparkOps[D <: DAO[_]](dao: D) {
```
pferrel, line 63: yes!

rest-server/core/src/main/scala/com/actionml/core/spark/SparkStoreSupport.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+import org.bson.codecs.configuration.CodecProvider
+
+import scala.reflect.ClassTag
+
+
+trait SparkStoreSupport {
+  def readRdd[T: ClassTag](
```
pferrel: Hmm, I assume there is some reason this has to be in a trait mixed-in to the `Store`. This should not be a public API but only used by the DAO's Spark support trait. Is there a way to protect access in some way?

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -270,7 +262,7 @@ class URAlgorithm private (engine: UREngine, initParams: String, dataset: URData
     }
     */
 
-    SparkContextSupport.getSparkContext(initParams, defaults).map { sc => // or implicit sc =>
+    SparkContextSupport.getSparkContext(initParams, engineId).map { sc => // or implicit sc =>
```
pferrel: Minor point -- it might be good to use the val name `appName = engineId` to self-document a bit.

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingAlgorithm.scala
pferrel: Oh, no. The collections should be in the Dataset, not created here. There should be no change in any use of the DAOs. The Store will not longer read a collection into an RDD, the DAO will. But this should have no effect on what class creates the DAO. You are correct, the Algorithm only manages the model, which we have not formalized yet. It may be a DAO but is certainly not the input DAO like the Events DAO, which is the responsibility of a Dataset.

pferrel:
One important comment about Algorithms not owning input DAOs. Those should be managed by the Dataset only. An Algorithm should create models from the Dataset, which should create and own all input DAOs. DAOs might be used in some future Engine to store models and therefor owned by an Algorithm but we do not have this now so the DAO should not be created in the Algo but kept in the Dataset as it is now.

pferrel:
This PR seem like 2 things, some cleanup of the getSparkContext, and the DAO.readRdd. Is this correct?

We still need to add reading multiple collections so All DAOs can have Spark support but we can do in another PR.

andreyshikov:
> We still need to add reading multiple collections so All DAOs can have Spark support but we can do in another PR.

I made a test and created 2 RDDs in one spark session, so it looks ok

PR #90, opened 2018-10-17
Feature/job manager

pferrel:


rest-server/core/src/main/scala/com/actionml/core/engine/LambdaAlgorithm.scala
```diff
@@ -30,7 +30,7 @@ trait LambdaAlgorithm[T] extends LazyLogging {
     * @param datum The input Event type defined by the parmeterized type
     * @return
     */
-  def input(datum: T): Validated[ValidateError, Boolean] = Valid(true)
+  def input(datum: T): Validated[ValidateError, String] = Valid("")
```
pferrel, line 33: Seems like we should 

- change all `Validated[ValidateError, Boolean]` cases to `Validated[ValidateError, String]`
- change ALL Strings returned to JSON

rest-server/core/src/main/scala/com/actionml/core/jobs/JobManager.scala
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/* Usage example
+val (jobId, future) = createJob(engineId = someString)
+future.map{ jobId =>
+  some code ...
+  removeJob(jobId)
+}
+Valid(s"jobId")
+*/
```
pferrel, line 60: seems to work

I realized that I don't need to manage a Queue since the ExecutionContext will do that so the `jobIds` are used only for reporting in status messages (not implemented yet).

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingAlgorithm.scala
```diff
@@ -238,8 +239,15 @@ class URNavHintingAlgorithm private (
...
-            logger.warn("Delete a \"user\" not supported")
-            Invalid(WrongParams("Using $delele on \"entityType\": \"user\" is not supported yet"))
+            //logger.warn("Delete a \"user\" not supported")
+            //Invalid(WrongParams("Using $delele on \"entityType\": \"user\" is not supported yet"))
+            val (jobId, future) = JobManager.createJob(engineId = engineId)
+            future.map{ _ =>
+              logger.info(s"Finished executing jobId: $jobId")
+              JobManager.removeJob(jobId)
+              logger.info(s"Active jobIds after finishing jobId: $jobId, are ${JobManager.getActiveJobIds(engineId)}")
+            }
```
pferrel, line 252: this Job is trivial just as an example of how it's used

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingDataset.scala
```diff
@@ -124,7 +124,7 @@ class URNavHintingDataset(engineId: String, val store: Store) extends Dataset[UR
           case "$delete" =>
             if (event.entityType == "user") {
               // this will only delete a user's data
-              itemsDao.removeOne(filter = ("entityId", event.entityId)) // removeOne all events by a user
+              //itemsDao.removeOne(filter = ("entityId", event.entityId)) // removeOne all events by a user
             } // ignore any other reserved event types, they will be caught by the Algorithm if at all
```
pferrel, line 129: this code is wrong. But in any case will be executed in a Job because it might take longer than the REST timeout allows.

pferrel:
comments on the "Files Changed"

pferrel:
Added a bunch of changes to return JSON for all response bodies.

pferrel:
ready to merge, the integration test fails due to actor system timeout during queries. This may be fixed by @andreyshikov in another feature branch via changes in the ESClient.

PR #92, opened 2018-11-05
Feature/mongo index management

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/store/Store.scala
```diff
@@ -19,11 +19,12 @@ package com.actionml.core.store
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
 
 
 trait Store {
   def dbName: String
-  def createDao[T](collectionName: String)(implicit ct: ClassTag[T]): DAO[T]
+  def createDao[T: TypeTag](collectionName: String)(implicit ct: ClassTag[T]): DAO[T]
```
pferrel, line 27: Why change from just [T] to a TypeTag?

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoStore.scala
```diff
@@ -18,25 +18,74 @@
...
 
 class MongoStorage(db: MongoDatabase, codecs: List[CodecProvider]) extends Store with LazyLogging {
-  import scala.concurrent.ExecutionContext.Implicits.global
   import MongoStorage.codecRegistry
 
-  override def createDao[T](name: String)(implicit ct: ClassTag[T]): DAO[T] = {
-    new MongoDao[T](db.getCollection[T](name).withCodecRegistry(codecRegistry(codecs)(ct)))
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  override def createDao[T: TypeTag](name: String)(implicit ct: ClassTag[T]): DAO[T] = {
```
pferrel, line 45: again, TypeTag? Is this required to have access to the annotations?

rest-server/core/src/main/scala/com/actionml/core/store/Store.scala
andreyshikov: because we need meta information - annotations

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoStore.scala
andreyshikov: yes

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoStore.scala
```diff
@@ -18,25 +18,74 @@
...
+    def getIndexNames: List[(String, Indexed)] = {
+      val symbol = symbolOf[T]
+      if (symbol.isClass) {
+        symbol.asClass.primaryConstructor.typeSignature.paramLists.head
+          .collect {
+            case f if f.annotations.nonEmpty && f.annotations.forall(_.tree.tpe =:= typeOf[Indexed]) =>
+              val i = f.annotations.head.tree.children.tail.foldLeft(Indexed()) {
+                case (acc, Select(t, n)) if t.tpe <:< typeOf[DurationConversions] =>
```
pferrel, line 54: Yikes, you need more comments. To me this is unreadable. Comments will at least give some hints.

rest-server/engines/src/main/scala/com/actionml/engines/urnavhinting/URNavHintingEngine.scala
```diff
@@ -138,11 +142,11 @@ object URNavHintingEngine {
       //eventId: String, // not used in Harness, but allowed for PIO compatibility
       event: String,
       entityType: String,
-      entityId: String,
+      @Indexed(order = asc) entityId: String,
       targetEntityId: Option[String] = None,
       properties: Map[String, Boolean] = Map.empty,
       conversionId: Option[String] = None, // only set when copying converted journey's where event = nav-event
-      eventTime: String) // ISO8601 date
+      @Indexed(order = desc, ttl = 30 days) eventTime: String) // ISO8601 date
```
pferrel, line 166: This should not be hard coded. We need to add a param in engine.json.

This might be a problem with annotations! This param is expected to change over time via updates to the engine.json. At first a user may want 30 days but later change it to 90 days with `harness update <some-engine-id>`

pferrel: @andreyshikov However, annotations will only affect the code when compiled--oops.

Only the TTL will change the index and order will never change and can be hard coded.

PR #106, opened 2018-12-13
added event aliases with integration tests

pferrel:
with and without event "aliases" seems to work the same way, as expected. The UR integration test now includes both with and w/o aliases

This is a required feature.

### DIFF
--- a/rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
+++ b/rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
@@ -106,7 +106,7 @@ class URAlgorithm private (
     var err: Validated[ValidateError, String] = Valid(jsonComment("URAlgorithm initialized"))
 
     recsModel = params.recsModel.getOrElse(DefaultURAlgoParams.RecsModel)
-    //val eventNames: Seq[String] = params.eventNames
+    //val indicatorParams: Seq[String] = params.indicatorParams
 
     userBias = params.userBias.getOrElse(1f)
     itemBias = params.itemBias.getOrElse(1f)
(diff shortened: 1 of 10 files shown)


rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -496,13 +497,29 @@ class URAlgorithm private (
 
     val userHistBias = query.userBias.getOrElse(userBias)
     val userEventsBoost = if (userHistBias > 0 && userHistBias != 1) Some(userHistBias) else None
+    // create a Map of alias -> indicator name or indicator name -> indicator name if no aliases
+    val queryEventNames = indicatorParams.flatMap { case (indicatorName, iParams) =>
+      val aliases = iParams.aliases.getOrElse(Seq(indicatorName))
+      if(indicatorName == aliases.head) {
```
andreyshikov, line 503: I'd suggest `aliases.contains(indicatorName)` or `alicases.headOption.contains(indicatorName` (if first position matters) to avoid NPE

rest-server/engines/src/main/scala/com/actionml/engines/ur/URPreparator.scala
```diff
@@ -46,23 +46,30 @@ object URPreparator extends LazyLogging with SparkMongoSupport {
...
-      eventNames = Some(eventNames),
+      indicatorParams = Some(indicatorParams),
       targetEntityType = Some(Some("item")))(sc).repartition(sc.defaultParallelism)
     */
 
     // now separate the events by event name
-    val eventRDDs: Seq[(ActionID, RDD[(UserID, ItemID)])] = eventNames.map { eventName =>
+    val indicatorNames = indicatorParams.keySet
+    val eventRDDs: Seq[(String, RDD[(UserID, ItemID)])] = indicatorNames.map { indicatorName =>
+      val aliases = indicatorParams(indicatorName).aliases.getOrElse(Seq(indicatorName))
```
andreyshikov, line 56: I'd suggest to use Map's getOrElse method `indicatorParams(indicatorName).getOrElse(aliases, Seq(indicatorName)`

qqmbr4k (approved):

PR #108, opened 2018-12-21
Fixed pretty print at JsonParser

qqmbr4k:


rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -155,8 +155,6 @@ class MongoAdministrator extends Administrator with JsonParser {
       }
     } else {
       logger.trace("Getting status for all Engines")
-      val statuses = jsonList(
-        engines.mapValues(_.status()).map(_._2.getOrElse(jsonComment(s"Warning: no status returned"))).toSeq)
```
pferrel, line 159: I don't think removing this is needed or good. This is where the status of engines is merged for display in the CLI

pferrel: What does removing this solve?

rest-server/server/src/main/scala/com/actionml/router/http/routes/EnginesRouter.scala
```diff
@@ -109,14 +109,14 @@ class EnginesRouter(implicit inj: Injector) extends BaseRouter with Authorizatio
   private def createEngine(implicit log: LoggingAdapter): Route = asJson { engineConfig =>
     log.info("Create engine: {}", engineConfig)
     completeByValidated(StatusCodes.Created) {
-      (engineService ? CreateEngine(engineConfig.toString)).mapTo[Response]
+      (engineService ? CreateEngine(engineConfig.noSpaces)).mapTo[Response]
     }
```
pferrel, line 113: In general all bodies returned to REST are formatted to be displayed even if compression would be better. This is because they are displayed by the CLI's use of the REST API. This may make the CLI responses unreadable--not sure?

pferrel: Ideally the API responses should be compressed and pretified in the Python CLI. But We can't compress without adding the python prettify or we'll have an unreadable CLI.

pferrel:
See comments on compressing JSON response bodies. These are in Validated strings. They are prettified now but ideally should be compressed for the REST API and prettified in the python CLI. I don't think we can do one without the other or we'll have unreadable CLI responses.

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
andreyshikov: @pferrel this variable is never used

rest-server/server/src/main/scala/com/actionml/router/http/routes/EnginesRouter.scala
andreyshikov: here is the input part, we will add pretty print on the output

pferrel: this PR must include the output, if you mean the python formatted response.

rest-server/core/src/main/scala/com/actionml/core/validate/JsonParser.scala
```diff
@@ -61,19 +63,12 @@ trait JsonParser extends LazyLogging {
   }
 
   def prettify(jsonString: String): String = {
```
andreyshikov, line 65: this is the output @pferrel 

> https://github.com/actionml/harness/pull/108#discussion_r243691750

PR #114, opened 2018-12-23
Added doc upsert for ES client

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/search/SearchSupport.scala
```diff
@@ -40,6 +40,7 @@ trait SearchClient[T] {
     fieldNames: List[String],
     typeMappings: Map[String, (String, Boolean)] = Map.empty,
     refresh: Boolean = false): Boolean
+  def saveOnById(id: String, typeName: String, doc: T): Boolean
```
pferrel, line 68: To make this work atomically there must be mapping generated for the fields of type `T`. Not sure this is required since mapping are determined when the writeRdd happens. However the relationship of T to ItemProperties to UREvent should be discussed. 

Data comes in via a `$set` UREvent, it is save as an ItemProperty to the `items` Mongo collection. It should also be saved in the ES model. Later when the model is re-generated the data from `items`will be processed into a model RDD which will create a new index. It is extremely important that the model created represents exactly what is in the `items` collection + the model data.

Is it true that the "upsert" will produce the equivalent of what is in the `events` collection? The type `T` seems to be the key. The structure of `ItemProperties` is not quite the structure of how they are saved in ES so we need to discuss how to solve the translation--I think.

andreyshikov: @pferrel yeah, let's discuss it by voice

PR #137, opened 2019-02-20
minor makefile improvments

dennybaa:


pferrel:
The test fails because you did not start from develop and modify develop. This is our preferred method to create PRs so the changed files will also be the only diff -- wherever possible.

pferrel:
If you think the test is wrong, please file an issue or ping someone to change it.

rest-server/bin/common.sh
```diff
@@ -13,22 +13,34 @@ export CYAN='\033[0;36m'         # Cyan--hints and info messages
 export WHITE='\033[0;37m'        # White
 
 
+## BSD/MacOS compatible readlink -f
```
dennybaa, line 16: @pferrel Here's the possible resolution for BSD/MacOS systems it resembles `readlink -f` behavior to locate HARNESS_HOME

Makefile
```diff
@@ -26,12 +26,13 @@ endif
...
 build: sbt
-	cd $(RESTSRV_DIR); \
+	cd $(RESTSRV_DIR) && \
 		$(SBT) ++$(SCALA_VERSION) -batch server/universal:stage
 
 clean: sbt
-	cd $(RESTSRV_DIR); \
+	cd $(RESTSRV_DIR) && \
 		$(SBT) ++$(SCALA_VERSION) -batch server/clean
+	find $(RESTSRV_DIR) -name "__pycache__" -exec rm -r {} \; || /bin/true
```
pferrel: @dennybaa  Doesn't bash have `true`? because I don't have `/bin/true` and have always used `true` in cases like this. From reading, `/bin/true` is linux/gnu specific

In any case I have to replace this to run it

dennybaa: As I'm aware `/bin/true` is *Nix specific.

Makefile
```diff
@@ -26,12 +26,13 @@ endif
...
 build: sbt
-	cd $(RESTSRV_DIR); \
+	cd $(RESTSRV_DIR) && \
 		$(SBT) ++$(SCALA_VERSION) -batch server/universal:stage
 
 clean: sbt
-	cd $(RESTSRV_DIR); \
+	cd $(RESTSRV_DIR) && \
 		$(SBT) ++$(SCALA_VERSION) -batch server/clean
+	find $(RESTSRV_DIR) -name "__pycache__" -exec rm -r {} \; || true
```
pferrel, line 35: changed `/bin/true` to which works on macOS and I think on all modern bash, at least I hope so.

rest-server/bin/common.sh
pferrel: Yes, this works! Thanks.

I found another way to replace many foundational commands with GNU versions but this seems somewhat dangerous since it is global and may break macOS specific scripts. I can also replace just `readlink` which seems like maybe a better idea since the BSD/masOS readlink is notoriously problematic.

Makefile
pferrel: and yet I don't have it. Does `true` not work someplace?

pferrel:
I had to change `/bin/true` in `Makefile` to `true` if anyone has trouble with this let me know

Makefile
dennybaa: Well hard to say need to try, since make uses sh not bash by default...

dennybaa:
I bet `|| true` should work in sh too

rest-server/bin/common.sh
dennybaa: Technically anyone can install GNU coreutils and on mac he will have greadlink binary, but we can not rely on this.

pferrel:
ok merging with develop

Thanks @dennybaa

PR #145, opened 2019-04-12
Added exception handler for DAO.saveOne

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -86,8 +86,14 @@ class URDataset(engineId: String, val store: Store) extends Dataset[UREvent](eng
       } // should be either aliases for an event name, defaulting to the event name itself
 
       if(aliases.contains(event.event)) { // only store the indicator events here
-        eventsDao.saveOne(event)
-        Valid(event)
+        try {
+          eventsDao.saveOne(event)
+          Valid(event)
+        } catch {
+          case e: Throwable =>
```
pferrel, line 93: @andreyshikov shouldn't this only catch the db exceptions?

PR #147, opened 2019-04-24
Feature/ttl config

pferrel:
calling code is fake, we only log the TTL to make sure it is parsed


rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoDao.scala
```diff
@@ -137,6 +157,34 @@ class MongoDao[T](val collection: MongoCollection[T])(implicit ct: ClassTag[T])
...
+          if (ttl.isFinite()) options.expireAfter(ttl.toMillis, TimeUnit.MILLISECONDS)
+          IndexModel(Document(iName -> 1), options)
+        }
+        if (newIndexes.nonEmpty) {
+          logger.info(s"Creating indexes ${newIndexes.map(i => s"${i.getKeys} - ${i.getOptions}")} for collection ${collection.namespace.getFullName}")
+          collection.createIndexes(newIndexes).toFuture
+        } else Future.successful(())
+      }.recover {
+      case e: Exception =>
+        logger.error(s"Can't create indexes for ${collection.namespace.getFullName}", e)
```
pferrel: @andreyshikov 

`case e: Exception` should be banned! This is dangerous to include everywhere when you actually mean a specific exception.

The more of these we have, the more likely a fatal exception or non-expected system level exception will be caught incorrectly. For instance an OOM should NOT be caught here.

I would expect you to correct me for this rather than the other way around :-)

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoStore.scala
```diff
@@ -43,57 +43,7 @@ class MongoStorage(db: MongoDatabase, codecs: List[CodecProvider]) extends Store
...
-          IndexModel(Document(iName -> 1), options)
-        }
-        if (newIndexes.nonEmpty) {
-          logger.info(s"Creating indexes ${newIndexes.map(i => s"${i.getKeys} - ${i.getOptions}")} for collection ${collection.namespace.getFullName}")
-          collection.createIndexes(newIndexes).toFuture
-        } else Future.successful(())
-      }.recover {
-      case e: Exception =>
-        logger.error(s"Can't create indexes for ${collection.namespace.getFullName}", e)
-    }
```
pferrel, line 96: Did you take out all annotated index hints?

Annotations indexing is required, the only case we have where an index may be created or modified at runtime is the TTL.

rest-server/engines/src/main/scala/com/actionml/engines/ur/UREngine.scala
```diff
@@ -141,7 +141,9 @@ object UREngine extends JsonSupport {
       mirrorType: Option[String] = None,
       mirrorContainer: Option[String] = None,
       sharedDBName: Option[String] = None,
-      sparkConf: Map[String, JValue])
+      sparkConf: Map[String, JValue],
+      algorithm: URAlgorithmParams,
```
pferrel, line 145: I thought about this and URAlgorithmParams is a requirement for this Engine. So this is correct

pferrel:
@andreyshikov can you review, respond and squash-merge this?

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoDao.scala
andreyshikov: OOM is an Error (subtype of Throwable), Exception can't be OOM or something that crashes jvm. ```case e => ``` is dangerous because it catches Throwables such as OOM error

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoStore.scala
andreyshikov: ok, I'll fix

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoDao.scala
pferrel: So `case e: Exception` is NOT dangerous and is considered good style?

andreyshikov: @pferrel I see no problems here with that code

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -106,7 +106,7 @@ class MongoAdministrator extends Administrator with JsonSupport {
     parseAndValidate[GenericEngineParams](json).andThen { params =>
       engines.get(params.engineId).map { existingEngine =>
         logger.trace(s"Re-initializing engine for resource-id: ${params.engineId} with new params $json")
-        enginesCollection.saveOne(EngineMetadata(params.engineId, params.engineFactory, json))
+        enginesCollection.saveOneById(existingEngine.engineId, EngineMetadata(params.engineId, params.engineFactory, json))
```
pferrel: @andreyshikov Are we deprecating `saveOne`? That would mean delete it from DAO. That would be ok with me.

rest-server/core/src/main/scala/com/actionml/core/store/Dao.scala
```diff
@@ -38,10 +39,11 @@ trait AsyncDao[T] {
   def insertManyAsync(c: Seq[T])(implicit ec: ExecutionContext): Future[Unit]
   def updateAsync(filter: (String, QueryCondition)*)(o: T)(implicit ec: ExecutionContext): Future[T]
   def saveOneByIdAsync(id: String, o: T)(implicit ec: ExecutionContext): Future[Unit]
-  def saveOneAsync(o: T)(implicit ec: ExecutionContext): Future[Unit]
```
pferrel, line 41: This API makes no sense. It seems good to remove. You cannot save without an ID or query.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -60,14 +67,28 @@ class URDataset(engineId: String, val store: Store) extends Dataset[UREvent](eng
...
+      parseAndValidate[URDatasetParams](
+        jsonConfig,
+        errorMsg = s"Error in the Dataset part pf the JSON config for engineId: $engineId, which is: " +
+          s"$jsonConfig",
+        transform = _ \ "dataset"
+      ).andThen { p =>
+        p.ttl.fold[Validated[ValidateError, _]](Valid(p)) { ttlString =>
+          Try(Duration(ttlString)).toOption.map { ttl =>
+            // We assume the DAO will check to see if this is a change and no do the reindex if not needed
+            eventsDao.createIndexes(Seq(("ttl_index", Indexed(ttl = ttl))))
```
pferrel: How is `ttl_index` mapped to a field in the stored case class? This should be a field name IMO, otherwise `ttl_index` is a magic value.

pferrel: Also the fold notation is overly hard to read.

Was there a logical error in the previous code?

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
andreyshikov: no, I changed savedOne syntax and now it looks good (it's update)

rest-server/core/src/main/scala/com/actionml/core/store/Dao.scala
andreyshikov: one can insert without id or query. def saveOne is useful

pferrel:
@andreyshikov is this finished?

qqmbr4k (approved):
looks ok for now

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
andreyshikov: @pferrel fixed that `ttl_index` bug, this fold is new

PR #149, opened 2019-05-19
Added compound indexes support for mongo

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/ur/UREngine.scala
```diff
@@ -157,25 +157,26 @@ object UREngine extends JsonSupport {
...
-      targetEntityId: Option[String] = None,
-      dateProps: Map[String, Date] = Map.empty,
-      categoricalProps: Map[String, Seq[String]] = Map.empty,
-      floatProps: Map[String, Float] = Map.empty,
-      booleanProps: Map[String, Boolean] = Map.empty,
-      @Indexed(order = desc, isTtl = true) eventTime: Date)
+                       eventId: Option[String], // not used in Harness, but allowed for PIO compatibility
+                       event: String,
+                       entityType: String,
+                       @SingleIndex(order = asc, isTtl = false) entityId: String,
```
pferrel, line 165: @andreyshikov does `false` have to be explicitly stated?

pferrel: also this seems to be a style problem here, indentation looks like Java style.

pferrel: in fact shouldn't order and isTtl be defaulted?

pferrel:
That was a big one!

PR #150, opened 2019-05-21
Feature/spark job server

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/jobs/JobManager.scala
```diff
@@ -45,87 +46,78 @@ trait JobManagerInterface {
...
           case (cancellable, _) =>
             removeJob(jobId)
             cancellable.cancel()
         }
-    }
-    Future.successful(())
+    } ++ Seq(LivyJobServerSupport.cancel(jobId)))
+      .map(_ => ())
+      .recover {
+        case e: Exception =>
```
pferrel: @andreyshikov I researched this and

- we should use the monad `Try` where possible because both options for completion do not catch fatal errors. Not applicable here
- in the case above we should use `NonFatal` NOT `Exception` here since some fatal errors are caught by `Exception`

rest-server/core/src/main/scala/com/actionml/core/jobs/JobManager.scala
```diff
@@ -45,87 +46,78 @@ trait JobManagerInterface {
...
   }
 }
 
-case class JobDescription(
-  jobId: String,
-  status: String = JobStatus.queued,
-  comment: String = "") extends Response
+case class JobDescription(jobId: String,
+                          status: JobStatus = JobStatuses.queued,
+                          comment: String = "") extends Response
```
pferrel: bad Scala style

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoDao.scala
```diff
@@ -31,19 +31,39 @@ import org.mongodb.scala.MongoCollection
...
+import org.bson.codecs.{Codec, DecoderContext, EncoderContext}
+import org.bson.{BsonReader, BsonWriter}
+import org.mongodb.scala.bson.collection.immutable.Document
+import org.mongodb.scala.model.IndexModel
+import org.mongodb.scala.{MongoClient, MongoCollection, MongoDatabase}
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
```
pferrel: @andreyshikov did you mean to import all these? I don't see a code change that would require them.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -294,21 +294,15 @@ class URAlgorithm private (
...
         // todo: for now ignore properties and only calc popularity, then save to ES
         calcAll(data, eventsRdd).save(dateNames, esIndex, esType, numESWriteConnections)
-
-        //sc.stop() // no more use of sc will be tolerated ;-)
       } catch {
-        case e: Throwable =>
-          logger.error(s"Spark computation failed for job $jobDescription", e)
-          sc.cancelJobGroup(jobDescription.jobId)
-          throw e
+        case e: Exception =>
```
pferrel: @andreyshikov again I think `NonFatal` is prefered and safer.

rest-server/server/src/main/resources/reference.conf
```diff
@@ -34,5 +34,6 @@ elasticsearch {
 }
 
 spark {
-  eventLog.dir="file:///tmp"
+  eventLog.dir = "file:///tmp"
+  job-server-url = ${SPARK_JOB_SERVER_URL}
```
pferrel, line 38: I would really like to stop using akka conf. If SPARK_JOB_SERVER_URL is changed elsewhere this might cause a bug. Why not just use the env to get these values instead of the syntactic sugar of akka conf. We know they have to be in the env anyway so this is not DRY to duplicate in akka conf.

pferrel (commented):
Some possible changes to review

rest-server/core/src/main/scala/com/actionml/core/jobs/JobManager.scala
andreyshikov: Try extracts NonFatal..., ok let's use NonFatal (at least it's a common Scala practice)

rest-server/core/src/main/scala/com/actionml/core/jobs/JobManager.scala
andreyshikov: let's use scalafmt, IDEA makes that ^^^ every time....

rest-server/server/src/main/resources/reference.conf
andreyshikov: spark.job-server-url is an app level setting, ${SPARK_JOB_SERVER_URL} is an env level one.
with akka.conf we have one place for mapping...

rest-server/core/src/main/scala/com/actionml/core/jobs/JobManager.scala
pferrel: it is a setting you can fix in IDEA. I don't bother with lots of coding conventions but we should stick to the ones in the style docs, right?

See the scala style docs. I have to fix these by hand every time I find them. I'll let these through. If you need help I can screen shot my settings, which are not perfect but catch this error at least.

rest-server/server/src/main/resources/reference.conf
pferrel: and it should be `harness-env` not .conf files spread around everywhere. Since we have to use harness-env it should be all we use, no?

pferrel: I see you point about being dry I still don't like having 2 mechanisms from reading env into the code. Why not ONLY have `harness-env`

pferrel:
Can you fix the conflict?

PR #151, opened 2019-05-27
Changed 'should' clause for ES6(7) query

qqmbr4k:


pferrel:
@andreyshikov does this work with ES5? Does it pass the integration test for any ES?

pferrel:
@andreyshikov I can't merge this? If you make a "Draft" PR please say why it is a draft.

andreyshikov:
> @andreyshikov does this work with ES5? Does it pass the integration test for any ES?

tested with 5, 6 and 7

pferrel:
@andreyshikov If we upgrade harness where there is already a server running ES 5.x, do we have a problem managing pre-existing indexes, which have a type of "item"???

rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
```diff
@@ -73,41 +76,40 @@ trait ElasticSearchResultTransformation extends JsonSearchResultTransformation[H
 class ElasticSearchClient[T] private (alias: String)(implicit w: Writer[T]) extends SearchClient[T] with LazyLogging with WriteToEsSupport with JsonSupport {
   this: JsonSearchResultTransformation[T] =>
   import ElasticSearchClient._
+  import ElasticSearchClient.ESVersions._
   implicit val _ = DefaultFormats
+  private val indexType = if (esVersion != v7) "items" else "_doc"
```
andreyshikov, line 81: > If we upgrade harness where there is already a server running ES 5.x, do we have a problem managing pre-existing indexes, which have a type of "item"???

it's not a harness problem, we use same "items" index type for 5 and 6, but we can't use it in 7. afaik upgrade from 5 to 6 should be quite easy. 6 to 7 - idk, reindexing maybe...

pferrel: I think this may solve my question, which is using this code with an old ES5 and index.

The case is 2 of our clients that have docker-compose with es5 and harness 0.4.0.

We would update all but ES5 unless they are ok with re-training. 

If they are ok with upgrading to ES7 and recreating the index it would not be an issue even without this change.

PR #153, opened 2019-06-07
Feature/json env

qqmbr4k:
Added support for ENV VARS links in JSON


rest-server/core/src/main/scala/com/actionml/core/validate/JsonSupport.scala
```diff
@@ -123,4 +128,12 @@ trait JsonSupport extends LazyLogging {
   }
 
 
+  private val env = s"system.env.(.*)".r
+  private def enrichViaSystemEnv: String => String = {
+    case env(value) => sys.env.getOrElse(value, {
+      logger.warn(s"No ENV VAR for $value")
+      value
+    })
```
andreyshikov, line 137: @pferrel this ^^^ means silence (just warning in the logs), is it ok?

pferrel: Sounds good.

pferrel (approved):
As I read this, there is no option for not enrichingViaSystemEnv. But we know the only valid use is for the engine JSON. I guess that is ok, there might be some as yet unforeseen use for Events or other JSON request bodies.

andreyshikov:
> As I read this, there is no option for not enrichingViaSystemEnv. But we know the only valid use is for the engine JSON. I guess that is ok, there might be some as yet unforeseen use for Events or other JSON request bodies.

@pferrel e.g. I can add a param (with default to "do not use system.env") to parseAndValidate to turn on "system.env parsing"

Makefile
```diff
@@ -34,7 +34,7 @@ clean: sbt
 		$(SBT) ++$(SCALA_VERSION) -batch server/clean
 	find $(RESTSRV_DIR) -name "__pycache__" -exec rm -r {} \; || true
 
-dist: clean clean-dist build wheel
```
andreyshikov: @pferrel that fixes the Makefile

PR #159, opened 2019-07-02
temp commit

qqmbr4k:


pferrel:
@andreyshikov is the same form of ES query used with ES5, 6, and 7? This should be possible.

pferrel:
@andreyshikov ok, got your answer on slack

merging

PR #162, opened 2019-07-04
Changed createIndexes API to support default TTL value (for db indexes)

qqmbr4k:


pferrel:
does it pass integration tests?

PR #165, opened 2019-07-22
Feature/fast import

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/backup/FSMirror.scala
```diff
@@ -129,4 +114,27 @@ class FSMirror(mirrorContainer: String, engineId: String)
     }
     Valid(jsonComment("Job created to import events in the background."))
   }
+
+  private def importFromFile(file: File, engine: Engine): Long = {
+    var eventsProcessed = 0
+    val src = Source.fromFile(file)
+    try {
+      src.getLines().sliding(512, 512).foreach { lines =>
```
andreyshikov, line 122: @pferrel do we need that setting?

pferrel: No idea. What is this setting based on? Does the `Source.fromFile` library not calculate a reasonable default? Did it speed up the import to set this `(512, 512)`???

@andreyshikov

rest-server/core/src/main/scala/com/actionml/core/engine/Engine.scala
```diff
@@ -126,9 +126,11 @@ abstract class Engine extends LazyLogging with JsonSupport {
   def input(json: String): Validated[ValidateError, Response] = {
     // flatten the event into one string per line as per Spark json collection spec
     mirroring.mirrorEvent(json.replace("\n", " ") + "\n")
-    Valid(Comment("Input processed by base Engine"))
+      .andThen(_ => Valid(Comment("Input processed by base Engine")))
   }
 
+  def inputMany: Seq[String] => Unit = _.foreach(input)
```
pferrel, line 132: Is this API extension important? It seems to only avoid one input per call. The call overhead should not be important, is it?

@andreyshikov

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -163,6 +153,44 @@ class URDataset(engineId: String, val store: Store) extends Dataset[UREvent](eng
     }
   }
 
+  override def inputMany(data: Seq[String]): Unit = {
+    val aliases = params.indicators.flatMap { ip =>
+      ip.aliases.getOrElse(Seq(ip.name))
+    }
+    val (events, items) = data.view.foldLeft[(List[UREvent], List[UREvent])]((List.empty, List.empty)) { case (acc@(el, il), jsonEvent) =>
```
pferrel, line 163: This is some pattern matching that I do not understand: `case (acc@(el, il)` what does it do? what is it called?

@andreyshikov

andreyshikov: I'll answer in slack

rest-server/core/src/main/scala/com/actionml/core/engine/Engine.scala
andreyshikov: It's not about call overhead, it's a sign that we can use a bulk insert that gives us a performance boost

pferrel:
@andreyshikov these conflicts need your resolution. these are the same as the conflicts in #169

PR #166, opened 2019-07-23
Added support for mongo uri

qqmbr4k:
Removed env vars MONGO_HOST, MONGO_PORT. Added env var MONGO_URI

### DIFF
--- a/rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoConfig.scala
+++ b/rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoConfig.scala
@@ -17,13 +17,16 @@
 
 package com.actionml.core.store.backends
 
+import com.mongodb.MongoClientURI
 import com.typesafe.config.ConfigFactory
 import net.ceedubs.ficus.Ficus._
 import net.ceedubs.ficus.readers.ArbitraryTypeReader._
 import net.ceedubs.ficus.readers.namemappers.implicits.hyphenCase
 
 
-case class MongoConfig(host: String, port: Int)
+case class MongoConfig(private val uri: String) {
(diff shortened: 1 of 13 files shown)


andreyshikov:
@pferrel tested with atlas, ready to merge

pferrel:
@andreyshikov This should be against `develop`

This requires changes to `harness-env` and perhaps the `harness-start` script to print the config for mongo that the server is using when started.

I do not see the env changes here.

pferrel:
Also we need docs for how to set the 2 users we need for Atlas.

rest-server/core/src/main/scala/com/actionml/core/store/backends/MongoConfig.scala
```diff
@@ -17,16 +17,14 @@
...
+case class MongoConfig(private val uriString: String) {
+  val uri = new MongoClientURI(uriString)
+}
 
 object MongoConfig {
-  private lazy val config = ConfigFactory.load
 
-  lazy val mongo = config.as[MongoConfig]("mongo")
-}
+  val mongo = MongoConfig(sys.env.getOrElse("MONGO_URI", throw new RuntimeException("Environment variable MONGO_URI must be set")))
```
andreyshikov, line 29: @pferrel am I right here?

pferrel: Looks correct, as long as we always set the value in script or code. If this exception is thrown it means someone has removed the setting.

pferrel:
this should be merged with `develop`, not the release branch.

@andreyshikov Will merge this branch by hand, not in the github gui.

PR #171, opened 2019-08-06
Feature/spark mongo auth develop

qqmbr4k:


pferrel:
Can't do this yet

We need to change the way the container manages env, it expects to see MONGO_HOST and MONGO_PORT, we need to change any docker-compose.yml files we maintain.

rest-server/bin/harness-env
```diff
@@ -8,9 +8,7 @@ export REST_SERVER_HOST=${REST_SERVER_HOST:-localhost}
 export REST_SERVER_PORT=${REST_SERVER_PORT:-9090}
 
 # MongoDB setup
-# todo: allow for multiple servers in cluster
-export MONGO_HOST=${MONGO_HOST:-localhost}
-export MONGO_PORT=${MONGO_PORT:-27017}
+export MONGO_URI=${MONGO_URI:-mongodb://localhost:27017}
```
pferrel, line 14: @andreyshikov ideally this creates MONGO_URI from MONGO_HOST and MONGO_PORT because these are passed into the container in the docker-composes we maintain along with k8s configs

Also if we want to pass in URIs we should probably do that everywhere and fix this only once in docker-compose and k8s

rest-server/bin/harness-env
```diff
@@ -8,9 +8,9 @@ export REST_SERVER_HOST=${REST_SERVER_HOST:-localhost}
 export REST_SERVER_PORT=${REST_SERVER_PORT:-9090}
 
 # MongoDB setup
-# todo: allow for multiple servers in cluster
 export MONGO_HOST=${MONGO_HOST:-localhost}
 export MONGO_PORT=${MONGO_PORT:-27017}
+export MONGO_URI=${MONGO_URI:-mongodb://${MONGO_HOST}:${MONGO_PORT}}
```
andreyshikov, line 13: RE: https://github.com/actionml/harness/pull/171/files#r311305505
@pferrel this way?

pferrel: @andreyshikov needed but not enough

To be truly backward compatible we need to then derive `MONGO_HOST` and `MONGO_PORT` from `MONGO_URI` that way no matter what a user sets in the env, all 3 vars will be set correctly based on whatever is supplied OR defaults.

specifically any docker-compose or k8s usage will work. Then docker-compose or k8s can upgrade as they have time based on the new use of URIs.

By this logic we should do this with ALL service pointers, ES, etc.

pferrel: I think we should just merge this and worry about full backward compatibility only if needed. Our containers may break but we can fix them.

PR #172, opened 2019-08-08
Feature/env refactor

pferrel:
@andreyshikov if you want to finish this here are some notes:

The merge of your mongo URI branch must have been done wrong. This branch is based on develop, then I included the version-cli branch and yours because we need to use URIs for all services.

To make this work we should extract nnn_HOST and nnn_PORT for MONGO, HARNESS, ELASTICSEARCH, etc AFTER the URI is constructed. This is to remain backward compatible but allow someone to set MONGO_URI and have MONGO_HOST and port set by the URI OR allow the user to set via the MONGO_URI exclusively. The reason for this backward compatibility is that containers already use the HOSTs and PORTs but we also want users to adopt the URIs for setting these. If they set via a URI in harness-env we should extract and put into MONGO_HOST etc. This requires parsing URIs in bash.

### DIFF
--- a/rest-server/bin/harness-env
+++ b/rest-server/bin/harness-env
@@ -8,9 +8,9 @@ export REST_SERVER_HOST=${REST_SERVER_HOST:-localhost}
 export REST_SERVER_PORT=${REST_SERVER_PORT:-9090}
 
 # MongoDB setup
-# todo: allow for multiple servers in cluster
 export MONGO_HOST=${MONGO_HOST:-localhost}
 export MONGO_PORT=${MONGO_PORT:-27017}
+export MONGO_URI=${MONGO_URI:-mongodb://${MONGO_HOST}:${MONGO_PORT}}
 
 # HDFS setup
 # This should be set by Hadoop, otherwise set it here
--- a/rest-server/bin/harness-env
+++ b/rest-server/bin/harness-env
@@ -8,7 +8,9 @@ export REST_SERVER_HOST=${REST_SERVER_HOST:-localhost}
 export REST_SERVER_PORT=${REST_SERVER_PORT:-9090}
 
 # MongoDB setup
-export MONGO_URI=${MONGO_URI:-mongodb://localhost:27017}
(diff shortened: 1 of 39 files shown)


rest-server/bin/harness-env
```diff
@@ -3,14 +3,16 @@
 # Harness Server config, should work as-id unless you are using SSL
 # to listen on the host IP address for external connections
 # export REST_SERVER_HOST=0.0.0.0 # to listen for external connections
-export REST_SERVER_HOST=${REST_SERVER_HOST:-localhost}
+export REST_SERVER_HOST=${REST_SERVER_HOST:-localhost} # BEING DEPRECATED, USE HARNESS_URI
 # to change the port used
-export REST_SERVER_PORT=${REST_SERVER_PORT:-9090}
+export REST_SERVER_PORT=${REST_SERVER_PORT:-9090} # BEING DEPRECATED, USE HARNESS_URI
+export REST_SERVER_PROTOCOL=${REST_SERVER_PROTOCOL:-http} # BEING DEPRECATED, USE HARNESS_URI
+export HARNESS_URI=${HARNESS_URI:-${REST_SERVER_PROTOCOL}://${REST_SERVER_HOST}:${REST_SERVER_PORT}} # NOT USED
```
pferrel, line 10: after setting HARNESS_URI we need to extract REST_SERVER_HOST from the URI that way users can set via REST_SERVER_HOST or HARNESS_URI and it will work.

Eventually we remove setting by REST_SERVER_HOST when our containers support the URIs

This should be done for allo services.

I have not included the parsing but some good scripts are here: https://stackoverflow.com/questions/6174220/parse-url-in-shell-script

pferrel:
This does not build because the feature/spark-mongo-auth-develop merge was not clean. @andreyshikov can you take a look at this?

We can't merge the URI support until the env refactoring is finished. And the version-cli only works if the URI support is included. So I've effectively made these branch's PRs dependent on this one.

pferrel:
this PR is obsolete

PR #175, opened 2019-09-05
Added logs and exception handling for insertMany and insert item

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -177,23 +177,30 @@ class URDataset(engineId: String, val store: Store) extends Dataset[UREvent](eng
           acc
       }
     }
-    eventsDao.insertMany(events)
+    try {
```
pferrel, line 180: @andreyshikov ideally shouldn't the DAO interface remove the use of exceptions?

andreyshikov: I don't want to ignore (just log) errors on DAO level - how can we e.g. respond to the client that something gone wrong?

PR #182, opened 2019-09-09
Feature/logger cleanup

pferrel:
Cleaning log messages

- in all cases it is know we need the engine-id in the message. Some errors may happen asynchronously so without the engine-id is is impossible to tell the affect of the error.
- we overuse `logger.info` and should use `logger.error` or `logger.warn` more often so log levels can be reduced without missing important messages

### DIFF
--- a/rest-server/conf/logback.xml
+++ b/rest-server/conf/logback.xml
@@ -9,18 +9,34 @@
 
     <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${HARNESS_LOG_PATH}/harness.log</file>
-        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
-            <!-- rollover daily -->
+        <rollingPolicy class="ch.qos.logback.core.rolling.FixedWindowRollingPolicy">
             <fileNamePattern>harness-%d{yyyy-MM-dd}.%i.log</fileNamePattern>
-            <!-- each file should be at most 1GB, keep 60 days worth of history, but at most 10GB -->
-            <maxFileSize>1GB</maxFileSize>
-            <maxHistory>60</maxHistory>
-            <totalSizeCap>10GB</totalSizeCap>
+            <minIndex>1</minIndex>
+            <maxIndex>51</maxIndex>
         </rollingPolicy>
+
+        <triggeringPolicy class="ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy">
+            <maxFileSize>1000KB</maxFileSize>
+        </triggeringPolicy>
         <encoder>
-            <pattern>%d{HH:mm:ss.SSS} %-5level %-17logger{0} - %msg%n</pattern>
+            <pattern>%date - [%level] - from %logger in %thread %n%message%n%xException%n</pattern>
         </encoder>
     </appender>
+
+    <!--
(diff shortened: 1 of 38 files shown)


pferrel:
not ready to merge, this is to keep track of ongoing logger message cleanup

rest-server/engines/src/main/scala/com/actionml/engines/ur/URPreparator.scala
```diff
@@ -36,56 +36,27 @@ object URPreparator extends LazyLogging {
...
-      targetEntityType = Some(Some("item")))(sc).repartition(sc.defaultParallelism)
-    */
-
     // now separate the events by event name
     val indicatorNames = indicatorParams.map(_.name)
 
     logger.info(s"Indicators: ${indicatorNames}")
 
-    //val collectedEventsRDD = eventsRDD.collect()
+    //val debug = eventsRDD.collect()
```
andreyshikov, line 44: ^

qqmbr4k (approved):

pferrel:
@andreyshikov this is a bunch of logger message cleanup. Conventions for messages:

- format of messages should include engine-id if present
- made as many info into trace as possible. Info should be things that are useful in a running system, trace are things that are usually not needed and may only show information about what harness or an engine is doing
- converted `debug` into `trace` for the most part

ready to merge

@andreyshikov you don't have to review all of these unless you want to.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URPreparator.scala
pferrel: @andreyshikov when you `.collect` any Spark data structure like an RDD it brings it all into memory in one process. This is ok for some debuging but very seldom used in real production Spark code. I have a convention of naming any collected thing "debug" so I can tell to not include them in production code. I would uncomment when using IntelliJ with small data. I have accidentally left these in when pushing so now I can just search for "debug" before pushing to make sure they are all taken out.

rest-server/conf/logback.xml
```diff
@@ -22,21 +23,20 @@
...
-    </rollingPolicy>
-    <encoder>
-        <pattern>%d{HH:mm:ss.SSS} %-5level %-17logger{0} - %msg%n</pattern>
-    </encoder>
-</appender>
--->
+    <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>${HARNESS_LOG_PATH}/harness.log</file>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <fileNamePattern>harness-%d{yyyy-MM-dd}.%i.log</fileNamePattern>
```
andreyshikov, line 31: yes, that's the place that I was talking about. This path is not the same as `${HARNESS_LOG_PATH}/harness.log`

qqmbr4k (changes requested):

rest-server/conf/logback.xml
pferrel: @andreyshikov this is the old config, yours is commented out.

If there is an error in this line what should it be?

andreyshikov: `${HARNESS_LOG_PATH}/harness-%d{yyyy-MM-dd}.%i.log` if you want them in the same dir

pferrel: aha, ok

PR #183, opened 2019-09-10
Job Status Removal

qqmbr4k:


pferrel:
Looks fine. 

@andreyshikov can you test the following:

1. submit many "import" commands to several engine-ids. Maybe a script with multiple import commands. This will add many commands to the queue but they are executed in parallel.
2. check to see they all are removed when completed 

I had trouble with this in a deployment and since all "jobs" are executed in parallel, it is a good case to test with the "job queue" idea.

PR #185, opened 2019-09-11
added harness-cli status system to CLI and the code to Harness, repor…

pferrel:
…ts build info and server config

When finished this will migrate to using URIs for all config in harness-env


rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
```diff
@@ -144,10 +144,20 @@ class MongoAdministrator extends Administrator with JsonSupport {
...
-    Valid(Comment(s"Version: ${com.actionml.admin.BuildInfo.version}"))
+    // todo: do we want to check connectons to services here?
+    Valid(SystemInfo(
+      buildVersion = com.actionml.admin.BuildInfo.version,
+      gitBranch = Properties.envOrElse("BRANCH", "No branch detected in env." ),
+      gitHash = Properties.envOrElse("GIT_HASH", "No short commit number detected in env." ),
+      harnessURI = Properties.envOrElse("HARNESS_URI", "No URI set, using host and port" ),
+      harnessHost = Properties.envOrElse("HARNESS_HOST", "No URI set, using host and port" ),
+      harnessPort = Properties.envOrElse("HARNESS_PORT", "No URI set, using host and port" ),
+      mongoURI = Properties.envOrElse("MONGO_URI", "No URI set" ),
```
andreyshikov: maybe it's time to put all those constants like "MONGO_URI" somewhere...

qqmbr4k (approved):

rest-server/bin/harness-start
```diff
@@ -13,10 +13,10 @@ HARNESS_CLASS=com.actionml.HarnessServer
...
 -Dlogback.configurationFile=${HARNESS_LOG_CONFIG} \
 -DHARNESS_LOG_PATH=${HARNESS_LOG_PATH} \
--Drest-server.host=${REST_SERVER_HOST} \
--Drest-server.port=${REST_SERVER_PORT} \
--Dmongo.host=${MONGO_HOST} \
--Dmongo.port=${MONGO_PORT}"
+-Drest-server.host=${HARNESS_HOST} \ # is this needed?
+-Drest-server.port=${HARNESS_PORT} \ # is this needed? We use a URI here
+-Dmongo.host=${MONGO_HOST} \ # is this needed?
+-Dmongo.port=${MONGO_PORT}" # is this needed? We use a URI here
```
pferrel: @andreyshikov are these really used? If so why do they have no effect on our switch to using the Mongo URI? I assume these actually set the values of akka conf values like `rest-server.host` and `mongo.port`

I also assume `mongo,host` is never used anymore. 

This is yet another reason we should completely remove akka conf. We must use the env for containers, this just makes problems like unused launch options more difficult to see. We need these in env, they should not be in launch options, they should be read from the env. Akka adds no value in our case, only not DRY setup.

pferrel: I believe these -D for config are not needed, is this correct @andreyshikov ?

rest-server/admin/src/main/scala/com/actionml/admin/MongoAdministrator.scala
pferrel: @andreyshikov yeah, I'm removing the akka conf for `rest-server` and deriving everything from one HARNESS_URI that can be set in the env. See changed to AppConfig here https://github.com/actionml/harness/pull/185/files#diff-5a7da7436100a71b67b47989b7df6de3R60

The removed akka conf is here: https://github.com/actionml/harness/pull/185/files#diff-fbbf7da5b09bc798c8bf296f595dc57dR9

pferrel: If this looks ok I'm going to move all addresses to URIs and remove some akka conf

PR #186, opened 2019-09-23
Integration test

tmaior:
Added integration test


pferrel:
we always squash-merge PR commits and the squashed comments are not very good as you can see.

PR #187, opened 2019-09-24
Added persistence for jobs

qqmbr4k:


pferrel (approved):
@andreyshikov 

Does this work correctly for:

 - harness restart, jobs are persistent but the Sparrk session will be restored on harness restart, right?
 - update, delete or the engine instance that started the job

In general job persistence may have a lot of side effects AND they will be different for import jobs and Spark training jobs.

PR #191, opened 2019-10-03
Added re-creation of UREngine indexes on engine update

qqmbr4k:


pferrel (changes requested):
Only new releases should be in PRs against master

this should go into develop or release/0.5.0 where we stage an 0.5.1 release not directly to master, right?

andreyshikov:
> Only new releases should be in PRs against master
> 
> this should go into develop or release/0.5.0 where we stage an 0.5.1 release not directly to master, right?

It's not needed in develop. If master doesn't need that change, let's delete this PR

PR #192, opened 2019-10-12
Added options for background index creation

qqmbr4k:
and  duplicated #191 into develop


rest-server/core/src/main/scala/com/actionml/core/engine/Engine.scala
```diff
@@ -47,7 +47,7 @@ abstract class Engine extends LazyLogging with JsonSupport {
     * @return The extending Engine instance, initialized and ready for input and/or queries etc.
     */
   @deprecated("Companion factory objects should be used instead of this old factory method", "0.3.0")
-  def initAndGet(json: String): Engine
+  def initAndGet(json: String, update: Boolean): Engine
```
pferrel, line 50: @andreyshikov this is under deprecation warning. Shouldn't we just remove it?

pferrel (approved):
I don't think initAndGet is being used. We use the companion object .apply now. 

@andreyshikov can you check this?

PR #193, opened 2019-10-16
Auto update images

tmaior:
Automatically updates harness on the K8s cluster with the new image version.

### DIFF
--- a/docker/config-dependencies.sh
+++ b/docker/config-dependencies.sh
@@ -10,16 +10,12 @@ curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s http
 chmod +x ./kubectl
 mv ./kubectl /usr/local/bin/kubectl
 
-
 # == Configure K8s Cluster
-# kubectl config set-credentials admin --username=${K8S_ADMIN_USER} --password=${K8S_ADMIN_PASS}
-# kubectl config set-cluster k8s.sitespect --insecure-skip-tls-verify=true --server=${K8S_SERVER}
-# kubectl config set-context k8s.sitespect --user=admin --namespace=default --cluster=k8s.sitespect
-# kubectl config use-context k8s.sitespect
+kubectl config set-credentials admin --username=${K8S_ADMIN_USER} --password=${K8S_ADMIN_PASS}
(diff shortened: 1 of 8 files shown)


docker/filetes.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+CLUSTER="devboar"
+
+if [[ "$CLUSTER" != "devboard" ]] && [[ "$CLUSTER" != "dev" ]] && [[ "$CLUSTER" != "devopsdev" ]]
```
pferrel, line 4: did you mean "devboar" as above?

This PR MUST be independent of which k8s cluster is ultimately going to deploy a new image

docker/filetes.sh
```diff
@@ -0,0 +1,10 @@
...
+      then
+         echo "sim"
+      else
+         echo "não"
```
pferrel, line 8: English please.

docker/config-dependencies.sh
```diff
@@ -0,0 +1,21 @@
...
+# == Install Kubectl
+curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
+chmod +x ./kubectl
+mv ./kubectl /usr/local/bin/kubectl
+
+# == Configure K8s Cluster
+kubectl config set-credentials admin --username=${K8S_ADMIN_USER} --password=${K8S_ADMIN_PASS}
+kubectl config set-cluster k8s.sitespect --insecure-skip-tls-verify=true --server=${K8S_SERVER}
+kubectl config set-context k8s.sitespect --user=admin --namespace=default --cluster=k8s.sitespect
+kubectl config use-context k8s.sitespect
```
pferrel: NO!!! Never include a client name in commit. This is open source code

pferrel:
@tmaior we MUST NEVER have client specific code in Harness! 

Auto updates must be driven from the k8s side, which IS client specific but still CANNOT have client names. If there is no way to do with without mentioning client names, we will not make the feature.

PR #194, opened 2019-10-18
Added ES index update

qqmbr4k:


pferrel:
Looks good afaict. 

I'll add a test for this in the integration test.

PR #196, opened 2019-10-21
Auto update image

tmaior:


pferrel:
@tmaior I don't understand how the re-deploy can be driven from harness, it may be deployed on many clusters. Doesn't it need to be a pull from the cluster?

pferrel:
@tmaior we need docs with this change

tmaior:
yeah @pferrel - I know. We will deliver this early next week. Had some issues this week.

PR #209, opened 2019-12-09
Release/0.5.1 rc1

pferrel:
merge with master and develop for release, do not merge until we decide to release


rest-server/server/src/main/resources/reference.conf
```diff
@@ -37,3 +37,7 @@ spark {
   eventLog.dir = "file:///tmp"
   job-server-url = ${SPARK_JOB_SERVER_URL}
 }
+
+jobs {
+  expire-after: ${JOBS_EXPIRE_AFTER}
```
pferrel, line 42: I thought we agreed to NOT use akka config and ConfigFactory. This is not DRY and has caused MANY bugs when we duplicate looking in the env in harness-env and in akka dot conf files.

Are we not on the same page here?

PR #214, opened 2020-01-07
Added JOBS_EXPIRE_AFTER var and increase training_sleep_seconds on build

tmaior:


.circleci/config.yml
```diff
@@ -77,10 +77,16 @@ jobs:
               docker-compose ps
               sleep 45
 
+              docker-compose exec harness-cli bash -c 'sed -i "s|training_sleep_seconds=40|training_sleep_seconds=60|g" /harness-cli/examples/ur/simple-integration-test.sh'
```
andreyshikov, line 80: @DioniPinho on my laptop 60 is not enough, I use 90. And resources that we use for ci for now is smaller then my laptop. I'd suggest something like 240 or 300

Also please take a look inside inner tests, they contain their own sleep_settings. I mean, looks like sed should be applied to e.g. `examples/ur/simple_test_ur_mobile_device_host_install.json`, not just `/harness-cli/examples/ur/simple-integration-test.sh`

qqmbr4k (changes requested):

qqmbr4k (approved):

PR #227, opened 2020-01-28
Added async findById for async query(_)

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
```diff
@@ -225,76 +226,66 @@ class ElasticSearchClient private (alias: String, client: RestClient)(implicit w
...
-        case 200 =>
-          val aliasResponse = client.performRequest(new Request("GET", s"/_alias/$alias"))
-          val responseJValue = parse(EntityUtils.toString(aliasResponse.getEntity))
-          val indexSet = responseJValue.extract[Map[String, JValue]].keys
-          rjv = indexSet.headOption.map { actualIndexName =>
-            val url = s"/$actualIndexName/$indexType/${encodeURIFragment(id)}"
-            logger.trace(s"Find doc by id using URL: $url")
-            val response = client.performRequest(new Request("GET", url))
-            logger.trace(s"Got response: $response")
+  def findDocById(id: String): EsDocument = Await.result(findDocByIdAsync(id), Duration.Inf)
```
pferrel, line 229: @andreyshikov Wouldn't a timeout be better? I mean Duration.Inf seems pretty long :-)

rest-server/core/src/main/scala/com/actionml/core/search/SearchSupport.scala
```diff
@@ -92,4 +92,5 @@ trait SearchClient[R, D] {
   def search(query: SearchQuery): Seq[R]
   def searchAsync(query: SearchQuery): Future[Seq[R]]
   def findDocById(id: String): D
+  def findDocByIdAsync(id: String): Future[D]
```
pferrel, line 95: @andreyshikov This makes me wonder if we should make create and delete index async since they could take a long time indeed. This is analogous to what you did to make changing the TTL async

rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
andreyshikov: > @andreyshikov Wouldn't a timeout be better? I mean Duration.Inf seems pretty long :-)

future will fail on at least ESclient timeout, and this timeout setting will be kind of duplication. I think it's not needed here...

PR #228, opened 2020-01-31
Created xmx var

tmaior:


rest-server/bin/harness-env
```diff
@@ -42,7 +42,7 @@ export ELASTICSEARCH_URI="${ELASTICSEARCH_URI:-http://localhost:9200}"
 export JOBS_EXPIRE_AFTER="12 hours"
 
 HARNESS_OPTS="\
--J-Xmx4G \
```
andreyshikov: afair we decided to set HARNESS_OPTS instead of it's parts (like HARNESS_XMX), but maybe I missed something... Anyway we need something we can pass as-is, e.g. for fine tuning I need to pass few settings for dispatcher configuration. So making a PR for each one doesn't look like a good idea.
Example of how many settings may be required https://doc.akka.io/docs/akka/current/general/configuration-reference.html

qqmbr4k (changes requested):

PR #231, opened 2020-02-12
Made mongo call from insert property synchronous

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -176,39 +176,38 @@ class URDataset(engineId: String, val store: Store) extends Dataset[UREvent](eng
...
-      _ <- itemsDao.saveOneByIdAsync(
-        event.entityId,
-        URItemProperties(
-          _id = updateItem._id,
-          dateProps = updateItem.dateProps ++ event.dateProps,
-          categoricalProps = updateItem.categoricalProps ++ event.categoricalProps,
-          floatProps = updateItem.floatProps ++ event.floatProps,
-          booleanProps = updateItem.booleanProps ++ event.booleanProps
-        )
+  private def insertProperty(event: UREvent): Unit = {
```
pferrel, line 179: Note: this should ONLY be called for a name: $set reserved event name. There should be no reason to filter out other event names by the time we are in this function.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -176,39 +176,38 @@ class URDataset(engineId: String, val store: Store) extends Dataset[UREvent](eng
...
-      }
-      esDoc = {
-        val doc = es.findDocById(event.entityId)
-        (doc._1, doc._2 ++ newProps)
-      }
-      result <- es.saveOneByIdAsync(event.entityId, esDoc)
-    } yield logger.trace(s"Document $esDoc ${if (result) " successfully saved" else " failed to save"} to Elastic Search")
-    ).recover {
-      case NonFatal(e) => logger.error(s"Engine-id: ${engineId}. Can't insert item $event", e)
+      case (name, _) => indicatorNames.contains(name)
```
pferrel, line 200: Note: this should ONLY be called for a name: $set reserved event name. There should be no reason to filter out other event names by the time we are in this function.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -176,39 +176,38 @@ class URDataset(engineId: String, val store: Store) extends Dataset[UREvent](eng
...
     }
+    (for {
+      doc <- es.findDocByIdAsync(event.entityId)
```
pferrel, line 203: this may not find a doc, it that ok here?

$set should create a mongo or es doc if one does not exist, or it should merge $set values with the existing doc. There is no failure mode for this unless the write operation fails.

pferrel:
@andreyshikov I'll merge for the experiment in staging but can you read the comments?

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
andreyshikov: this `findDocByIdAsync` method return empty result if it can't find by entityId

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
andreyshikov: yes, this method is called above only for $set items

rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
andreyshikov: there is a verification above (check that it's $set and item)

PR #233, opened 2020-02-13
Feature/make insert property mongo call sync

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/ur/URDataset.scala
```diff
@@ -191,22 +191,18 @@ class URDataset(engineId: String, val store: Store) extends Dataset[UREvent](eng
...
         event.booleanProps.mapValues(_.toString :: Nil) ++
         event.dateProps.mapValues { d => writeFormat.format(d.toInstant) :: Nil } ++
         event.floatProps.mapValues(_.toString :: Nil)
-      ).filterNot {
-      case (name, _) => indicatorNames.contains(name)
+      ).filterNot { case (name, _) => indicatorNames.contains(name) }
+    es.saveOneByIdAsync(event.entityId, esDoc).map { result =>
+      logger.trace(s"Document $esDoc ${if (result) " successfully saved" else " failed to save"} to Elastic Search")
+    }.onFailure {
+      case NonFatal(e) => logger.error(s"Engine-id: ${engineId}. Can't insert item $event", e)
```
pferrel: @andreyshikov Can we give more information? Do we know this fails due to non-existent index? Is this where a failure would happen for no index? A no-index situation should not cause a Harness failure response, only a Mongo failure should cause that or a failure for ES when the index exists. See comments in Slack.

BTW should the $event be in the log message not $event.entityId?

PR #242, opened 2020-03-29
Bugfix/popular eventnames

pferrel:


pferrel:
Still needs an update to harness-cli to reflect the new correct results in the popularity model and needs the training fix merged.

PR #243, opened 2020-03-30
Fixed ES update error for  events

qqmbr4k:


pferrel:
@andreyshikov does this include the training fix?

rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
```diff
@@ -128,9 +125,10 @@ class ElasticSearchClient private (alias: String, client: RestClient)(implicit w
 
   def saveOneByIdAsync(id: String, doc: EsDocument): Future[Comment] = {
     val promise = Promise[Comment]()
-    try {
-      val request = new Request("POST", s"/$alias/$indexType/${encodeURIFragment(id)}/_update")
-      request.setJsonEntity(
+    val msg404 = "The Elasticsearch index does not exist, have you trained yet?"
+    def sendUpdate(indexName: String) = {
+      val updateRequest = new Request("POST", s"/$indexName/$indexType/${encodeURIFragment(id)}/_update")
```
pferrel, line 130: The code here: `Request("POST", s"/$indexName/$indexType/${encodeURIFragment(id)}/_update")` is different than above in saveOneById, which is: `Request("POST", s"/$alias/$indexType/${encodeURIFragment(id)}/_update")`

Are these both correct?

pferrel: This is a request that must be sent with an indexName, not an alias? If so it might cause race conditions. The alias gave us one atomic action that would update the index for every consumer.

For instance this might change a value in an index as another thread is changing the alias and destroying the index that was upserted into.

Any time we update an index directly rather than going through an alias is therefor suspect.

pferrel (approved):
OK for now but still have the alias and URI questions.

rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
andreyshikov: fixed, thanks

PR #244, opened 2020-04-03
Feature/distributed engines

qqmbr4k:


rest-server/admin/src/main/scala/com/actionml/admin/Administrator.scala
```diff
@@ -17,28 +17,188 @@
...
+case class EnginesStatuses(statuses: List[Response]) extends Response
+
+case class SystemInfo(
+                       buildVersion: String,
+                       gitBranch: String,
+                       gitHash: String,
+                       harnessURI: String,
+                       mongoURI: String,
+                       elasticsearchURI: String
+                     ) extends Response
```
pferrel: indent wrong

rest-server/core/src/main/scala/com/actionml/core/config/AppConfig.scala
```diff
@@ -15,46 +15,64 @@
...
+case class RestServerConfig(host: String, port: Int, sslEnabled: Boolean)
+case class ActorSystemConfig(name: String, timeout: FiniteDuration)
+
 case class AuthConfig(enabled: Boolean,
-  serverUrl: String,
-  clientId: String,
-  clientSecret: String)
+                      serverUrl: String,
+                      clientId: String,
+                      clientSecret: String)
```
pferrel: indent

rest-server/core/src/main/scala/com/actionml/core/engine/Engine.scala
```diff
@@ -122,9 +124,9 @@ abstract class Engine extends LazyLogging with JsonSupport {
     * todo: can we combine the json output so this can be inherited to supply status for the data the Engine class
     * manages and the extending Engine adds json to give stats about the data it manages?
     */
-  def status(): Validated[ValidateError, Response] = {
+  def status(): HIO[Response] = {
     logger.trace(s"Status of base Engine with engineId:$engineId")
-    Valid(EngineStatus(engineId, "This Engine does not implement the status API"))
+    IO.succeed(EngineStatus(engineId, "This Engine does not implement the status API"))
```
pferrel, line 129: We are COMPLETELY replacing Validated with HIO?

This is a major change. Validated is used everywhere.

rest-server/core/src/main/scala/com/actionml/core/engine/backend/EngineMetadata.scala
```diff
@@ -14,13 +14,17 @@
...
-package com.actionml.admin
+package com.actionml.core.engine.backend
 
 import org.bson.codecs.configuration.CodecProvider
 
-object MongoStorageHelper {
 
+case class EngineMetadata(engineId: String,
+                          engineFactory: String,
+                          params: String)
```
pferrel: indent

rest-server/core/src/main/scala/com/actionml/core/validate/JsonSupport.scala
```diff
@@ -104,6 +106,29 @@ trait JsonSupport extends LazyLogging {
     }
   }
 
+  def parseAndValidateIO[T](json: String,
+                            errorMsg: String = "",
+                            transform: JValue => JValue = a => a)(implicit tag: TypeTag[T], ct: ClassTag[T], mf: Manifest[T]): HIO[T] = {
```
pferrel: indent

rest-server/server/src/main/resources/reference.conf
```diff
@@ -4,6 +4,8 @@ include "akka-ssl.conf"
 
 actor-system {
   name = "harness-router"
+  timeout = 5 seconds
+  timeout = ${?HARNESS_HTTP_TIMEOUT}
```
pferrel, line 8: why is this changed?

rest-server/server/src/main/resources/reference.conf
```diff
@@ -35,10 +37,16 @@ auth {
...
+  job-server-url = "http://localhost:8998"
+  job-server-url = ${?SPARK_JOB_SERVER_URL}
 }
 
 jobs {
   expire-after: 12 hours
   expire-after: ${?JOBS_EXPIRE_AFTER}
 }
+
+etcd-config {
```
pferrel, line 49: We agreed to NEVER add to settings in akka configs. They must ALL be duplicated in harness-env so they are not needed in akka config.

Right?

The exception is where a unit test needs config, right?

rest-server/server/src/main/resources/reference.conf
andreyshikov: this setting correlates with `etcd-config.timeout` and should be greater to complete the http response

rest-server/server/src/main/resources/reference.conf
andreyshikov: Yes, we agreed. But few weeks ago we discussed it again and there was a case with debugging and (maybe) running tests in IDEA, and this is exact case. I don't think this var should be set in harness-env.

rest-server/core/src/main/scala/com/actionml/core/engine/Engine.scala
andreyshikov: This change is like replacing `Validated[E,A]` with `Future[Validated[E,A]]`. It's required to make things "async" (put calculation inside a container)

rest-server/server/src/main/resources/reference.conf
pferrel: This should be removed from akka conf file and put in harness-env as a regular env var. We agreed not to duplicate env in akka conf.

rest-server/server/src/main/resources/reference.conf
pferrel: there are other ways to support env in IDEA that I use for all the other env. I really do not like duplicating this config.

The only thing we NEED to do in akka is conf for unit tests -- IIRC

But I will resolve for now.

PR #245, opened 2020-04-03
Fixed indexType (and Elasticsearch version) initialization

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
```diff
@@ -88,6 +88,20 @@ class ElasticSearchClient private (alias: String, client: RestClient)(implicit w
...
   implicit val _ = DefaultFormats
+  private val esVersion: ESVersions.Value = {
+    import ESVersions._
+    Try {
+      val response = client.performRequest(new Request("GET", "/"))
+      val version = (JsonMethods.parse(response.getEntity.getContent) \ "version" \ "number").as[String]
+      if (version.startsWith("5.")) v5
+      else if (version.startsWith("6.")) v6
+      else if (version.startsWith("7.")) v7
+      else {
```
pferrel, line 99: @andreyshikov actually I think we can assume v8+ will work untill we find otherwise.

pferrel:
We will release this to master through the normal release process making it a new version. Closing.

PR #247, opened 2020-04-04
Bugfix/re enable mirroring

pferrel:
re-enabling mirroring


rest-server/core/src/main/scala/com/actionml/core/backup/FSMirror.scala
```diff
@@ -39,6 +39,7 @@ class FSMirror(mirrorContainer: String, engineId: String)
   // java.io.IOException could be thrown here in case of system errors
   override def mirrorEvent(json: String): Validated[ValidateError, String] = {
```
pferrel, line 40: @andreyshikov should this be done in a thread for the case of `engine.inputAsync` ???

PR #252, opened 2020-05-21
Accelerated mirroring on import

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/ur/UREngine.scala
```diff
@@ -137,10 +137,7 @@ class UREngine extends Engine with JsonSupport {
       Future.successful(response)
   }
 
-  override def inputMany(data: Seq[String]): Unit = {
-    super.inputMany(data)
```
pferrel, line 141: @andreyshikov 

mirroring SHOULD happen on import. 

Some users do not use the REST api to input, they create files periodically and import them one after the other over time. So import should behave exactly like `input` . I think this means `inputMany` SHOULD call `super.input`

This will cause a problem -- infinite loop -- if you import from the mirror location and this is documented.

pferrel: ok merging but we should look at this again to clean up mirroring and add the "export" feature.

PR #253, opened 2020-05-29
Feature/es7 by default

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
```diff
@@ -96,12 +96,12 @@ class ElasticSearchClient private (alias: String, client: RestClient)(implicit w
       else if (version.startsWith("7.")) v7
       else {
         logger.warn("Unsupported Elastic Search version: $version")
-        v5
+        v7
       }
     } catch {
       case NonFatal(e) =>
         logger.error("Elasticsearch version can't be determined. v5 will be used.", e)
```
pferrel: @andreyshikov is this the wrong logger.error

rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
```diff
@@ -96,12 +96,12 @@ class ElasticSearchClient private (alias: String, client: RestClient)(implicit w
       else if (version.startsWith("7.")) v7
       else {
```
pferrel, line 97: @andreyshikov assume we are using ES8, this should not be a WARN, it should be assumed to work fine.

andreyshikov:
@pferrel all changes from the comments ^^ applied. I'd like to merge, if you don't mind

PR #255, opened 2020-07-08
added feature to use the item-based query for an itemSet with only on…

pferrel:
…e item

### DIFF
--- a/rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
+++ b/rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
@@ -567,11 +567,10 @@ class URAlgorithm private (
         params.itemSetBias))
       )
     } else { // one item in the set so better to use an item query
-      val newBias = if(query.itemSetBias.nonEmpty) query.itemSetBias else params.itemSetBias
       getSimilarItemsMatchers(
         query.copy(
           item = query.itemSet.get.headOption,
-          itemBias = newBias
+          itemBias = query.itemSetBias.orElse(params.itemSetBias)
         )
       )
     }
--- a/rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
+++ b/rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
@@ -561,7 +561,7 @@ class URAlgorithm private (
 
   private def getItemSetMatchers(query: URQuery)(implicit ec: ExecutionContext): Future[Seq[Matcher]] = {
     if(query.itemSet.getOrElse(Seq.empty).size != 1){
-      Future(query.itemSet.getOrElse(Seq.empty).map(item => Matcher(
+      Future.successful(query.itemSet.getOrElse(Seq.empty).map(item => Matcher(
         modelEventNames.head, // only look for items that have been converted on
         query.itemSet.getOrElse(Seq.empty),
         params.itemSetBias))
--- a/rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
+++ b/rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
@@ -462,11 +462,11 @@ class URAlgorithm private (
     val numResults = query.num.getOrElse(limit)
 
     // create a list of all query correlators that can have a bias (boost or filter) attached
-    getUserHistMatcher(query).zip(getSimilarItemsMatchers(query)).map { case ((userHistoryMatchers, userEvents), similarItemsMatchers) =>
+    getUserHistMatcher(query).zip(getSimilarItemsMatchers(query)).zip(getItemSetMatchers(query)).map { case (((userHistoryMatchers, userEvents), similarItemsMatchers), itemSetMatchers ) =>
       val shouldMatchers =
         userHistoryMatchers ++
         similarItemsMatchers ++
-        getItemSetMatchers(query) ++
+        itemSetMatchers ++
         getBoostedRulesMatchers(aggregatedRules)
 
       val mustMatchers = getIncludeRulesMatchers(aggregatedRules)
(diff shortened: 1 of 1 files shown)


rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -462,11 +462,11 @@ class URAlgorithm private (
     val numResults = query.num.getOrElse(limit)
 
     // create a list of all query correlators that can have a bias (boost or filter) attached
-    getUserHistMatcher(query).zip(getSimilarItemsMatchers(query)).map { case ((userHistoryMatchers, userEvents), similarItemsMatchers) =>
+    getUserHistMatcher(query).zip(getSimilarItemsMatchers(query)).zip(getItemSetMatchers(query)).map { case (((userHistoryMatchers, userEvents), similarItemsMatchers), itemSetMatchers ) =>
```
pferrel, line 465: Is this the correct way to include getItemSetMatchers now that the function returns a Future[Seq[Matchers]]? I had to change the return value of this function so that I could use getSimilarItemsMatchers for the case when the query has only one item in the itemSet.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -558,11 +558,23 @@ class URAlgorithm private (
...
+  private def getItemSetMatchers(query: URQuery)(implicit ec: ExecutionContext): Future[Seq[Matcher]] = {
+    if(query.itemSet.getOrElse(Seq.empty).size != 1){
+      Future(query.itemSet.getOrElse(Seq.empty).map(item => Matcher(
+        modelEventNames.head, // only look for items that have been converted on
+        query.itemSet.getOrElse(Seq.empty),
+        params.itemSetBias))
+      )
+    } else { // one item in the set so better to use an item query
+      val newBias = if(query.itemSetBias.nonEmpty) query.itemSetBias else params.itemSetBias
+      getSimilarItemsMatchers(
```
pferrel, line 570: switching to getSimilarItemsMatchers when only one item in the itemSet. Had to make getItemSetMatchers return Future[Seq[Matcher]]  so that both paths return the same thing, namely a Future. No need for a Future in the "if" part but should not hurt anything either, right?

pferrel:
Not sure if this passes test yet. Will be testing July 8

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -558,11 +558,23 @@ class URAlgorithm private (
...
 
-  private def getItemSetMatchers(query: URQuery): Seq[Matcher] = {
-    query.itemSet.getOrElse(Seq.empty).map(item => Matcher(
-      modelEventNames.head, // only look for items that have been converted on
-      query.itemSet.getOrElse(Seq.empty),
-      params.itemSetBias))
+
+  private def getItemSetMatchers(query: URQuery)(implicit ec: ExecutionContext): Future[Seq[Matcher]] = {
+    if(query.itemSet.getOrElse(Seq.empty).size != 1){
+      Future(query.itemSet.getOrElse(Seq.empty).map(item => Matcher(
```
andreyshikov: I'd suggest to use `Future.successful` instead of `Future` (Future.apply indeed) because this code only creates case class instance, so no need to do this concurrently in the separate thread.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
andreyshikov: yes, but I'd suggest to use Future.successful (see comment above)

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
andreyshikov: yes, you can add it here. In that case two [possible] getSimilarItemsMatchers queries to ES will be executed concurrently. If you want them to be executed sequentially, you can use flatMap instead of zip or a for-comprehension. Something like
```
    val userHistMatcherF = getUserHistMatcher(query)
    val similarItemsMatcherF = getSimilarItemsMatchers(query)
    val itemSetMatcherF = getItemSetMatchers(query)
    for {
      (userHistoryMatchers, userEvents) <- userHistMatcherF
      similarItemsMatchers <- similarItemsMatcherF
      itemSetMatchers <- itemSetMatcherF
      mustMatchers = getIncludeRulesMatchers(aggregatedRules)
      mustNotMatchers = getExcludeRulesMatchers(aggregatedRules) ++ getBlacklistedItemsMatchers(query, userEvents)
      shouldMatchers = userHistoryMatchers ++
        similarItemsMatchers ++
        itemSetMatchers ++
        getBoostedRulesMatchers(aggregatedRules)
    } yield SearchQuery(
      sortBy = rankingsParams.head.name.getOrElse("popRank"), // todo: this should be a list of ranking rules
      should = shouldMatchers,
      must = mustMatchers,
      mustNot = mustNotMatchers,
      filters = getDateFilters(query),
      size = numResults,
      from = startPos
    )
```
maybe you'll find this more readable, looks ok to me in any form.

qqmbr4k (changes requested):
lgtm but please replace `Future`[apply] with `Future.successful` in the `getItemSetMatchers` method

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
```diff
@@ -558,11 +558,23 @@ class URAlgorithm private (
...
+        modelEventNames.head, // only look for items that have been converted on
+        query.itemSet.getOrElse(Seq.empty),
+        params.itemSetBias))
+      )
+    } else { // one item in the set so better to use an item query
+      val newBias = if(query.itemSetBias.nonEmpty) query.itemSetBias else params.itemSetBias
```
andreyshikov: this line is an equivalent of ` val newBias = query.itemSetBias.orElse(params.itemSetBias)`

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
pferrel: Concurrently is exactly what we want, item and itemSet queries are possible but are labeled "experimental" and so not really supported. Therefore we expect only one of user, item, itemSet but it should work as expected if all or some are in the query.

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
pferrel: yes, thankyou

rest-server/engines/src/main/scala/com/actionml/engines/ur/URAlgorithm.scala
pferrel: So Future.successful will return a completed Future of status "successful"? I have read a lot on Futures but this is the first time I have seen this explained in a simple understandable way. So Future.successful is a way to fool the type system into treating a non-concurrent bit of code the same as a multi-threaded bit, right?

pferrel: Aha, I found what you are saying in a blog: "The Future.successful method constructs an already completed Future with the specified result. " 

This explains a lot and is correct here.

qqmbr4k (approved):

PR #290, opened 2021-03-08
Feature/distributed engines v2

qqmbr4k:
Refactored engines management to use only 2 entities (engines and harness-instances, dropped "actions")


rest-server/admin/src/main/scala/com/actionml/admin/Administrator.scala
```diff
@@ -173,10 +169,11 @@ trait Administrator extends LazyLogging with JsonSupport {
...
-      logger.trace(s"Getting status for $resourceId")
-      engines(resourceId).status()
-    } else {
+    engines.collectFirst {
+      case (meta, e) if meta.engineId == resourceId =>
+        logger.trace(s"Getting status for $resourceId")
+        e.status()
+    }.getOrElse {
       logger.error(s"Non-existent engine-id: $resourceId")
       IO.fail(WrongParams(jsonComment(s"Non-existent engine-id: $resourceId")))
```
pferrel, line 178: @pferrel should we have a better error

rest-server/core/src/main/scala/com/actionml/core/engine/backend/EnginesEtcdBackend.scala
```diff
@@ -17,264 +17,221 @@
...
-        w
-      }}.onError(_.failureOption.fold(IO.unit)(w => IO.effect(Try(w.close())).ignore))
-        .mapErrorCause(_ => Cause.fail(ValidRequestExecutionError()))
+      harnessIdQ <- Queue.bounded[Long](1)
+      actions <- kv.get(enginesPrefix, kvPrefixOpt(enginesPrefix))
+                   .map(_.getKvs.asScala.flatMap(s => decode(s.getValue)))
+      _ = actions.foreach {
+        case a: Add => callback(a)
+        case u: Update => callback(u)
+        case _ =>
```
pferrel, line 97: how does deleting and engine work?

PR #291, opened 2021-03-09
Fixed UR#train method for ES7 API change

qqmbr4k:
Fixes the situation when harness does not delete old indexes after training on ES7


pferrel:
What can cause indexes to NOT be deleted?

PR #300, opened 2021-04-22
Enabled int-test for all branches

DioniPinho:
Enabled int-test for all branches, but publish images of specific branches only


pferrel:
@DioniPinho Does this split the old "publish" into "test" and "publish"? Where test is run always, and publish only if test passes and we are on develop, master, or devops (or experimental)?

pferrel:
@DioniPinho It looks like this leaves the phases: build, publish

since we do not publish on most branches should there be: build, test, publish?

PR #306, opened 2021-06-13
Added user-delete and user-get methods

qqmbr4k:


rest-server/core/src/main/scala/com/actionml/core/search/SearchSupport.scala
```diff
@@ -93,4 +93,5 @@ trait SearchClient[R, D] {
   def searchAsync(query: SearchQuery): Future[Seq[R]]
   def findDocById(id: String): D
   def findDocByIdAsync(id: String): Future[D]
+  def deleteDoc(query: SearchQuery): Future[Unit]
```
pferrel, line 96: neither delete nor get should affect ES data.

rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
```diff
@@ -329,6 +328,20 @@ class ElasticSearchClient private (alias: String, client: RestClient)(implicit w
     promise.future
   }
 
+  override def deleteDoc(query: SearchQuery): Future[Unit] = {
```
pferrel, line 331: When is this used? I don't think GET or DELETE should affect ES data.

rest-server/server/src/main/scala/com/actionml/router/http/routes/EnginesRouter.scala
```diff
@@ -160,4 +166,18 @@ class EnginesRouter(implicit inj: Injector) extends BaseRouter with Authorizatio
       (engineService ? CancelJob(engineId, jobId)).mapTo[Validated[ValidateError, Response]]
     }
   }
+
+  private def getUserData(engineId: String, userId: String) = parameters('num.as[Int].?, 'from.as[Int].?) { (num, from) =>
```
pferrel, line 170: We should rename any reference to "user" into "entity". Harness should not recognize USER objects, this is a concept not found in Harness or even the UR. For instance the REST is GET harness/engines/engine-id/entities/entity-id

rest-server/server/src/main/scala/com/actionml/router/service/EngineService.scala
```diff
@@ -69,6 +77,30 @@ class EngineServiceImpl(implicit inj: Injector) extends EngineService{
...
+      } { engine =>
+        sender() ! (try {
+          engine.getUserData(userId, num, from)
+        } catch {
+          case _: NotImplementedError => Invalid(NotImplemented)
+          case NonFatal(_) => Invalid(ValidRequestExecutionError)
+        })
+      }
+
+    case DeleteUserData(engineId, userId) =>
```
pferrel, line 93: userId should be "entityId"

see other comments about the term "user"

rest-server/core/src/main/scala/com/actionml/core/search/elasticsearch/ElasticSearchSupport.scala
pferrel: not used? Ok.

PR #309, opened 2021-07-13
Fixed 'properties' field in getEvent response

qqmbr4k:


rest-server/engines/src/main/scala/com/actionml/engines/ur/UREngine.scala
```diff
@@ -235,6 +235,29 @@ object UREngine extends JsonSupport {
...
+  ) extends Response
+  object UREventOriginal {
+    def fromUREvent(e: UREvent): UREventOriginal =
+      UREventOriginal(
+        e.eventId,
+        e.event,
+        e.entityType,
+        e.entityId,
+        e.targetEntityId,
+        properties = e.dateProps ++ e.floatProps ++ e.booleanProps ++ e.categoricalProps,
```
pferrel, line 256: have you tried this with actual properties? I'm unsure if this creates the correct format.

In our case we should never have properties so I'll approve but this should at least be commented with a "todo:" 
as untested.

andreyshikov: yes, I tried, all looks good - {}

